PR #12, opened 2015-10-02
Tap :reference now means something. fixes #10

### LINKED ISSUE #10: tap/:reference not respected
The tap endpoint [accepts a `reference` argument](https://github.com/facebook/WebDriverAgent/blob/c6c4ebdf664731145dee4f2738bd35529ee06d8f/WebDriverAgentLib/WebDriverCommands/FBElementCommands.m#L35), but does nothing with it. I assume the idea is to have that be an element id, and to tap relative to that.

jaredly:
I wasn't sure what the desired way to say "no reference" is. If you give
something that's not a number, then it will treat that as "reference from the
top left of the screen".

See #10


mmmulani:
awesome!

PR #13, opened 2015-10-05
Guard against nil dictionary insertion for FBCapabilitiesCommands

lawrencelomax:
Some of these can be nil in iOS 7 so its best to avoid crashing the server and just let the client figure out what it should do with missing values.


mmmulani:
lgtm

PR #14, opened 2015-10-07
Use Helper method to allocate port

lawrencelomax:
I created this a while back, but forgot to wire it up 😣

Adds some additional error info on the startup path.


mmmulani:
lgtm

PR #15, opened 2015-10-14
Concrete Object for Sync/Async Routes

lawrencelomax:
The overwhelming majority of the responses to WebDriver commands are synchronous, which makes returning the result via a callback an unnecessary indirection. It also means that Always-Be-Returning(™)-Early requires the callback to be called, then `return`.

To keep the routing succinct, there's a concrete object that contains all of the information about how to handle a route (verb/path/handler) as opposed to encoding this information in a string as well as a key-value pair in a dictionary. 

This will also make it possible to build paths from a 'Session Route' do remove the need to include `:session` in all the paths. 

//cc @asm89 @mmmulani


mmmulani:
lgtm

PR #19, opened 2015-10-14
Implement Element Caching

lawrencelomax:
Element Caching is unimplemented and will always return a new index, regardless of the same element being fetched.

This PR uses `-[UIElement uiaxElement]` as a unique value. `NSMapTable` weaking on the `UIAXElements` is used to ensure that the elements can be released when they are deallocated elsewhere.


mmmulani:
lgtm

PR #27, opened 2015-10-23
log a message when the server starts

### LINKED ISSUE #25: Output message when server is ready to accept requests
I have a test which programmatically launches WebDriverAgent on a simulator and hits the /tree endpoint after the WebDriverAgent process outputs its first message (`Built at XX:XX:XX`). This test fails.

But if I wait for 3 seconds before issuing the http request, it passes.

I think we should output a message when the http server is ready. I suggest something similar to appium like:
`WebDriverAgent REST http interface listener started on 0.0.0.0:8100`

Jonahss:
fixes #25


Jonahss:
*signed cla

PR #29, opened 2015-10-26
Session ID validation on all relevant requests

lawrencelomax:
This PR tidies up a lot of the routing behavior with the ultimate aim of introducing a common way of:
- Checking for the existence of session ids on any relevant request.
- A trapdoor for disabling the validation of sessions
- Allowing 'sessionless' requests such as `/tree/` and `/status/` that don't need an active session in order to execute.

It does this by providing switches on `FBRoute` that do the right thing in terms of chaining response-handling and route-registration together.

There's also an update to a `GET` on `/session/<session-id>/` where just the capabilities, but after reading the spec I'm unsure of whether this endpoint should return the session id alongside or nested in the capabilities.


marekcirkos:
Closed in favour of https://github.com/facebook/WebDriverAgent/pull/70

PR #38, opened 2015-10-31
Name

Jonahss:
I'm thinking that the facebook tests must be sending the element id in the url *and* the message body, since commands like `/value` and `/clear` work for you.

`/enabled` returned a string ("0" or "1") instead of a boolean. Spec requires a boolean though.

Are these changes going to break your tests? How should we handle breaking changes to the current api like this?


mmmulani:
it might break our tests but we our infra should catch it, so I'm okay with it :)

PR #41, opened 2015-11-06
Fixed orientation command

marekcirkos:


marekcirkos:
Pulled more changes then I intended :)

PR #39, opened 2015-11-06
Extracted core WDA functionality to enable code reuse for UIAutomatio…

marekcirkos:
…n WDA and XCTest WDA

PR #43, opened 2015-11-07
Xcui test

Jonahss:


Jonahss:
this mainly catches up the XCUITest branch to the stuff that was merged into master

marekcirkos:
Nice work! Unfortunately XCTest is limited to just one simulator test run, so
we decided to merge UIAutomation and XCTest approach into one project (separate targets) and support both. This will prevent two branches from driving into different direction and help to benefit from improvement in either of them. So after merging https://github.com/facebook/WebDriverAgent/pull/39 I'll create new XCUITest branch & PR to add support for XCUITest. I'll also add your improvements, but in case I miss something let me know.

Jonahss:
Aha yes I see. It can only run one simulator at a time because xcodebuild
launches the sims.
I’ve noticed that if I run xcodebuild with the -destination "id=$SIM_UDID"
argument, if the sim specified is already booted xcodebuild will connect to
the already-running sim.
I’ve been able to run in parallel on multiple real-devices this way.
Have you tried starting multiple sims using FBSimulatorControl and then
running xcodebuild for each? This might skip the xcodebuild code path that
kills all running simulators and boots a new one.

This move still works in our favor because it allows for running tests on
versions of iOS less than 9.0.
​

marekcirkos:
Have a look https://github.com/facebook/WebDriverAgent/tree/XCUITest-WDACore

PR #44, opened 2015-11-09
Added FBSession class that handles creating and fetching session rela…

marekcirkos:
…ted object, like cache etc. Due to nature of WDA test there is restriction to only one session at the time.


mmmulani:
lgtm

PR #45, opened 2015-11-10
Added WebDriverAgent that works with XCUITest framework

marekcirkos:


ericwengelking:
@marekcirkos Is there any documentation on how to run WebDriverAgent using FBXCTWebDriverAgent instead of FBUIAWebDriverAgent?  By following the instructions in the Readme.md file I can get WebDriverAgent running with FBUIAWebDriverAgent.  I can run a command such as http://localhost:8100/tree with the UIA agent.  I'd like to start exploring what commands are available and what responses look like with the FBXCTWebDriverAgent.  Any help would be appreciated.

marekcirkos:
Sorry for that, will add docs soon. Till then you can use xcodebuild or Xcode to start XCTUITestRunner by running test in UITestingUITests. It will start simulator with running  WebDriverAgent. All you need to do is start new session. To do this check this command https://github.com/facebook/WebDriverAgent/blob/XCUITest-WDACore/XCTWebDriverAgentLib/Commands/FBSessionCommands.m#L26

ericwengelking:
For me when I run the tests the XCTStubApp crashes in main.m line 15. 

```obj-c
int main(int argc, char * argv[])
{
  @autoreleasepool {
    return UIApplicationMain(argc, argv, nil, NSStringFromClass([FBUIAppDelegate class]));
  }
}
```
I'm using Xcode 7.1 with an iOS 9.1 6s simulator.

marekcirkos:
@ericwengelking I updated README. Are still having an issue starting XCTWebDriverAgent?

mmmulani:
mega change but lgtm!

ericwengelking:
No longer having issues after pulling the latest from master @marekcirkos .  Thanks for this and the updated Readme!

PR #46, opened 2015-11-11
change desired capability names to comply with conventions/spec

Jonahss:


Jonahss:
also /name endpoint (tagName)

XCTWebDriverAgentLib/Commands/FBSessionCommands.m
```diff
@@ -26,10 +26,10 @@ + (NSArray *)routes
     [[FBRoute POST:@"/session"] respond:^ id<FBResponsePayload> (FBRouteRequest *request) {
 
       NSDictionary *requirements = request.arguments[@"desiredCapabilities"];
-      NSString *bundleID = requirements[@"CFBundleIdentifier"];
-      NSString *appPath = requirements[@"FBBinaryPath"];
-      NSAssert(bundleID, @"Should have bundle ID");
-      NSAssert(appPath, @"Should have app path");
+        NSString *bundleID = requirements[@"bundleId"];
+        NSString *appPath = requirements[@"app"];
```
marekcirkos: Isn't this too ambiguous. Maybe `appPath` would be better?

Jonahss: I agree that appPath is better, but the spec would have to be changed.
The reason is that with `app` you can specify a path to an app file, a url
to download the app from, or the bundleId of the app already installed on
the device.
https://code.google.com/p/selenium/source/browse/spec-draft.md?repo=mobile

PR #50, opened 2015-11-12
Size

Jonahss:


WebDriverAgentLib/Categories/UIAElement+WebDriverAttributes.m
```diff
@@ -37,6 +37,17 @@ - (NSDictionary *)wdRect
...
 
+- (NSDictionary *)wdSize
+{
+    CGRect rect = self.wdFrame;
+    return
+    @{
+        @"width": @(rect.size.width),
+        @"height": @(rect.size.height),
+      };
+
```
marekcirkos: Whitespace

XCTWebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -61,7 +66,7 @@ + (NSArray *)routes
       NSInteger elementID = [request.parameters[@"id"] integerValue];
       XCUIElement *element = [elementCache elementForIndex:elementID];
       id text;
-      if ([element elementType] == XCUIElementTypeStaticText) {
+      if ([element elementType] == XCUIElementTypeStaticText || [element elementType] == XCUIElementTypeButton) {
```
marekcirkos: Alternative is to create wdText property, but this is only command we need it for so I'll leave it as is.

PR #51, opened 2015-11-12
/back

Jonahss:


mmmulani:
this command seems a bit weird. it doesn't fully replicate the back button on Android or the navigation idea of "back".

what do we lose by not supporting it? (like, do you have any numbers on how many tests use it?)

Jonahss:
No such numbers exist :(

It's one of those "convenience" methods that appium can provide just for better device-parity. I think the method for finding it here is reliable, though of course a user could replicate the functionality themselves.

I'm ok with leaving it out for now, if you think it isn't necessary.

lawrencelomax:
@Jonahss since `WebDriverAgent` in general is a good place to run code on a device, it might be feasible to provide some sort of plugin architecture to allow loading of Framework bundles at runtime. I'm not sure how feasible this is on a Device with codesigning, but it may be possible to load a signed bundle since we don't have to rely on App-Store review rules.

mmmulani:
yeah I would leave it out for now. let's wait for some user reports and
figure the right way to implement this

Jonahss:
I can always just implement it in the appium xcuitest driver.

mmmulani:
closing this as per our discussion

PR #52, opened 2015-11-13
Location

Jonahss:


XCTWebDriverAgentLib/Categories/XCUIElement+WebDriverAttributes.m
```diff
@@ -108,4 +108,13 @@ - (NSDictionary *)wdSize
       };
 }
 
+- (NSDictionary *)wdLocation
+{
+    return
```
marekcirkos: For `wdSize ` and `wdLocation` I think it would be better to reuse `wdRect` by calling `return self.wdRect[@"size"]` and `return self.wdRect[@"origin"]`

Jonahss: :+1: I'll make the change

Jonahss:
closed in favor of #58

PR #55, opened 2015-11-17
Apple tv osx targets

marekcirkos:
This adds limited support for OS X and AppleTV. I think it is worth to pull it for the project structure.


marekcirkos:
Will leave it as playground branch

PR #56, opened 2015-11-18
By.id() == By.name()

lawrencelomax:
By.name() is the best proxy that we currently have for By.id(). Clients that attempt to use By.id() will get confusing behavior if this search method is unsupported.


marekcirkos:
lgtm

PR #59, opened 2015-11-20
add Pods directly to project, so users can install off npm without ru…

Jonahss:
…nning cocoapods

I understand if this isn't your normal way of using cocoapods, but this way users will be able to install the entire project through npm and dependencies will come with it.

We want appium users to be unaware of the fact that they're building an xcode project. We should also periodically update the pods.

If you don't want to include them this way, could we keep this in a supported branch? And keep rebasing off master. I can use the branch for publishing appium drivers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,8 @@ xcuserdata
 # you should judge for yourself, the pros and cons are mentioned at:
 # http://guides.cocoapods.org/using/using-cocoapods.html#should-i-check-the-pods-directory-into-source-control
 #
-Pods/
+
+# Pods/
 
 # Carthage
 #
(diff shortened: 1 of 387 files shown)


mmmulani:
we wanted to add these originally and I checked my email thread internally with legal and unfortunately, we aren't allowed to distribute these :( (I guess it's because people might think that Facebook wrote it or is providing it? even though we obviously aren't)

I'm not sure even a branch would be allowed although a fork (that wasn't Facebook owned) would obviously be okay. sorry for the trouble :(

Jonahss:
Aha I see. Yeah that totally makes sense.
Ok, I'll set up a branch that just includes this commit. Makes the release process a little messier but I can cope :)

I'll set it up once I'm ready to publish the first version in appium, then I'll come back here and close this PR.

Jonahss:
Forked with Pods bundled: https://github.com/appium/WebDriverAgent

PR #67, opened 2015-11-30
Fixed alert accept & dismiss commands. 

marekcirkos:
Calling [XCUICoordinate tap] have some side effect which looks like additional tap on the screen. Switching to XCEventGenerator which is used by -[XCUIElement tap]


mmmulani:
lgtm

PR #71, opened 2015-12-02
Added inhibit_all_warnings! to Podfile

marekcirkos:
Worthless warnings


lawrencelomax:
Sounds good, I'm not sure if WDA has this but it might be nice to use an `xcconfig` to provide a good default set of warnings. I've tripped over a few things in `FBSimulatorControl` as I didn't have enough warnings set at the time.

PR #76, opened 2015-12-11
Preventing XCUITestRunner from stopping on app crash or other failure

marekcirkos:
At the moment whenever application crashes during test or we encounter any other failure test runner is immediately stopped as well.
It is way better to have test runner still responding to requests infroming that application is dead and possibly serve other requests.


Jonahss:
cool!

PR #92, opened 2016-01-08
Adding support for Touch ID Match and NoMatch

kaleposhobios:
BTouchIDCommands contains new route /simulator/touch_id with argument “match” to trigger Touch ID Match or NoMatch event. If you need me to rebase my changes better I can.

I have signed the corporate CLA.

Issue #26, opened 2015-10-22
Simulate fingerprint ID on simulators?


Jonahss:
According to this, looks like there exists a framework for simulating touch ID: http://stackoverflow.com/questions/25459059/is-there-any-way-to-use-apples-touch-id-fingerprint-scanner-on-ios-simulator

Think we could find the way to call it ourselves?

mmmulani:
it looks like that posts a notification of name "com.apple.BiometricKit_Sim.fingerTouch.match"/"com.apple.BiometricKit_Sim.fingerTouch.nomatch". If that's enough, it should be pretty easy but haven't tested it out. Do you want to try it?

Jonahss:
I've got a couple higher-priority things first I wanna clear out. Working on getting a working webdriveragent-driver for appium. Then I'm going to start looking for unimplemented api endpoints.

mmmulani:
sounds good :)

-Mehdi

Jonahss:
Could you explain how you detected that notification?

mmmulani:
I just opened the simulator in Hopper and started poking around for some
touch ID stuff

kaleposhobios:
I tested this and it seems to work. I added a route like this:
    [[FBRoute POST:@"/touch_id_fail"] respond: ^ id<FBResponsePayload> (FBRouteRequest *request) {
        if (notify_post("com.apple.BiometricKit_Sim.fingerTouch.nomatch")) {
            return FBResponseDictionaryWithOK();
        } else {
            return FBResponseDictionaryWithStatus(FBCommandStatusUnsupported, nil);
        }
    }],
I tested it on my touch id enabled app and it works great. 

I'm not sure which route file you want to add this to, but I confirm the notification you mentioned will work and we can easily implement it.

mmmulani:
I would add a new FBTouchIDCommands.m file here: https://github.com/facebook/WebDriverAgent/tree/master/WebDriverAgentLib/WebDriverCommands and add them

Jonahss:
In appium we have custom routes like this:
`'/wd/hub/session/:sessionId/appium/device/lock'`

You could add a route like
`/session/:sessionId/xcuitest/simulator/touch_id`. Pass a boolean parameter for pass/fail? Or something along those lines.

kaleposhobios:
Why WebDriverAgentLib and not XCTWebDriverAgentLib? Or to put it another way, do I have to duplicate the code? Can I simply add the file to both targets?

[...]

PR #93, opened 2016-01-11
Added Simulator Touch ID Match and NoMatch commands

kaleposhobios:
FBTouchIDCommands contains new route /simulator/touch_id with argument “match” to trigger Touch ID Match or NoMatch event


kaleposhobios:
Note: this is a re-working of the previous PR https://github.com/facebook/WebDriverAgent/pull/92, so its cleaner and doesn't have Pods.

marekcirkos:
I see you added it to UIAutomation WebDriverAgent does it work only on UIAutomation?
If it works for both UIAutomation and XCUITest could you move files to 'WebDriverAgentCore/Commands'?

kaleposhobios:
Hi,

Thanks for taking a look. I didn't actually try it on UIAutomation, only on XCUITest. I put it in the location as suggested by mmmulani on this thread: https://github.com/facebook/WebDriverAgent/issues/26 but probably I didn't accurately communicate which mechanism it works with.

Since it's a just a notification posted to the system I'll move it to ```WebDriverAgentCore/Commands```

kaleposhobios:
Ok, I have moved it. Please let me know if there's anything else I can do.

marekcirkos:
Awesome, thank you!

marekcirkos:
I needed to revert your diff due to issues with our, bot. They have been merged back here https://github.com/facebook/WebDriverAgent/commit/fa239bdea378b1030627d95f8e8fe60e69c19863

PR #99, opened 2016-01-15
HACK: Redefine free() to fix bug in AXRuntime.framework

mmmulani:


Podfile
```diff
@@ -28,11 +31,13 @@ end
 # XCT Testing
 
 target :XCTWebDriverAgentLib, :exclusive => true do
+  pod 'fishhook', :git => 'https://github.com/facebook/fishhook.git', :commit => '8dbd09b'
```
marekcirkos: Do we needed in XCT as well?

mmmulani: I'm not sure but I think so because we share the same main.m? was more afraid of breaking it to be honest :P

marekcirkos:
This insanity looks good to me :)
image

Jonahss:
Mad props

PR #100, opened 2016-01-15
Removed all .gitattributes files, as they are not needed

marekcirkos:


marekcirkos:
@facebook-github-bot shipit

PR #113, opened 2016-01-21
Returning error description instead of NSError

marekcirkos:


marekcirkos:
@facebook-github-bot shipit

Issue #86, opened 2016-01-04
Can we use this to run XCUITests for an app ?


gop1334:
Currently we have set of XCUITests written in Swift for our app and we execute them on simulator. If FBSimulatorControl can launch parallel simulators in the same machine, can we use XCTWebDriverAgent to run the tests in parallel in the simulators launched ? Is this doable and if yes how can one do it ? Any information on this would very helpful. Thanks!

marekcirkos:
If you got XCUITests written in swift you don't need XCTWebDriverAgent. XCTWebDriverAgent uses XCUITest to test/automate applications by http instructions (more details check https://code.google.com/p/selenium/wiki/WebDriverForMobileBrowsers). 

Unfortunately you can't run multiple XCUITest in parallel with FBSimulatorControl yet, but we are working on it as we need this for XCTWebDriverAgent as well.

lawrencelomax:
> can we use XCTWebDriverAgent to run the tests in parallel in the simulators launched?

It's not currently possible to do this since there appears to be a daemon process `testmanagerd` running on the mac that prevents multiple tests running concurrently for the current user. It is however possible to setup a simulator using `FBSimulatorControl` and inspect it after launching with `DVTOperation`.

However, being able to run tests in parallel will become more important as `XCTWebDriverAgent` matures.

Jonahss:
Multiple real-devices works though, since they each choose a different port
:)

bootstraponline:
What about refactoring the lib so it can be used by people writing XCTest directly? Features like `XCUIElement+FBIsVisible.h` are generally useful regardless if webdriver is used or not. It'd expand the potential contributor base to include everyone using XCTest.

Jonahss:
Kinda of like a separate framework, something like FacebookXCTestUtils?

bootstraponline:
Yep.

marekcirkos:
I don't think we have anything useful except for `scrollToVisible`, which I bet Apple will implement soon. Also instead of using `isFBVisible` you can simply use Apple's `isHittable`. We use `isFBVisible`, because we need it on `FBElementSnapshot`, which is faster to construct element tree etc.

bootstraponline:
> I don't think we have anything useful except for scrollToVisible, which I bet Apple will implement soon.

Pretty much all the `XCTWebDriverAgentLib` commands are useful. Given the state of XCTest, I don't think waiting for Apple to implement things is realistic. 

[...]

PR #117, opened 2016-02-05
Add XCTHelper framework

bootstraponline:
#86

I'm new to Objective C so if there are any issues, let me know.

The code compiles. I don't think there are E2E tests to verify the helper code works.


marekcirkos:
I don't think it make sense to extract everything to separate framework. Extracting 'utilities' to separate classes within WebDriverAgent library would be enough, as linker will pull only used classes.

bootstraponline:
Ok, will update.

bootstraponline:
I'll submit a new PR once updated.

marekcirkos:
I'll be moving most of the code in that manner for better testability, so you can wait or help :)

bootstraponline:
Cool. I might wait then. I'm currently working on some non-automation projects and will be back to iOS in a few weeks.

PR #122, opened 2016-03-12
Added hide keyboard command

kaleposhobios:
Adds a new command and route for XCTest to hide the keyboard.

### DIFF
--- a/XCTWebDriverAgentLib/Commands/FBCustomCommands.m
+++ b/XCTWebDriverAgentLib/Commands/FBCustomCommands.m
@@ -25,6 +25,7 @@
   @[
     [[FBRoute POST:@"/deactivateApp"] respondWithTarget:self action:@selector(handleDeactivateAppCommand:)],
     [[FBRoute POST:@"/timeouts/implicit_wait"] respondWithTarget:self action:@selector(handleImplicitWaitCommand:)],
+    [[FBRoute POST:@"/hide_keyboard"] respondWithTarget:self action:@selector(handleHideKeyboard:)]
   ];
 }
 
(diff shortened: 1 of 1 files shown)


XCTWebDriverAgentLib/Commands/FBCustomCommands.m
```diff
@@ -51,4 +52,11 @@ + (NSArray *)routes
   return FBResponseDictionaryWithOK();
 }
 
++ (id<FBResponsePayload>)handleHideKeyboard:(FBRouteRequest *)request
+{
+    FBXCTSession *session = (FBXCTSession *)[FBSession activeSession];
```
marekcirkos: Could you use session from request property?

XCTWebDriverAgentLib/Commands/FBCustomCommands.m
```diff
@@ -51,4 +52,11 @@ + (NSArray *)routes
...
+    [[session.application.windows elementBoundByIndex:0] tap];
```
marekcirkos: Could you verify if element you are tapping is expected one? Otherwise it might introduces hard to spot issues

marekcirkos:
Thank you for PR!

XCTWebDriverAgentLib/Commands/FBCustomCommands.m
kaleposhobios: @marekcirkos Thanks for this. Right now it's actually just grabbing the Window and calling tap on it. My assumption was it's hitting it at (0,0) but I don't see in the docs for sure where it would hit, the docs just say "a hittable point". 

I'm sure this is not the ideal way, but based on what Apple is easily giving us, as a (hopefully) short-term solution to easily close the keyboard, we've just been doing it this way in our tests.

Are you ok with this?  I can definitely change it if desired. Let me know what you think.

XCTWebDriverAgentLib/Commands/FBCustomCommands.m
kaleposhobios: Will do.

XCTWebDriverAgentLib/Commands/FBCustomCommands.m
marekcirkos: I understand that, however you could check if keyboard is even visible before performing tap.

It is interesting it tap on (0, 0). I would expect it taps and the centre, anyway. :)

kaleposhobios:
@marekcirkos Thanks. I added a check as suggested. Also I rebased to get it onto the latest commit in your repo. Please let me know what you think.

marekcirkos:
@facebook-github-bot shipit

marekcirkos:
Awesome! Thank you for PR.

marekcirkos:
@facebook-github-bot shipit

PR #125, opened 2016-04-05
Fix travis for UIAWebDriverAgent

marekcirkos:
Linking UIKit since Travis build is failing due to missing UIDevice symbols.


marekcirkos:
@facebook-github-bot shipit

PR #126, opened 2016-04-05
Fix readme commands

bootstraponline:
The readme commands are wrong...

- `carthage bootstrap` builds all platforms (including tvOS). I think WDA supports only iOS?
- `WebDriverAgent.xcworkspace` doesn't exist anymore due to the move to carthage. Also `id='<DEVICE_UDID>'` is pretty unfriendly syntax since no one knows their UDIDs. 
- `Failed to authorize rights (0x1) with status: -60007.` happens if dev mode is disabled yet the readme lacks the fix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,16 +10,27 @@ If you are looking for WebDriverAgent that uses `UIAutomation.framework` check [
 Our dependencies are tracked with Carthage. First run
 
 ``
-carthage bootstrap
+carthage bootstrap --platform ios
 ``
 
 and then open `WebDriverAgent.xcodeproj`.
 
 ## Running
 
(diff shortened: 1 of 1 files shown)


marekcirkos:
I missed that, thank you for the fixes. Could you also update readme about `Failed to authorize rights (0x1) with status: -60007`.

bootstraponline:
sure, updated

bootstraponline:
github has a very big outage right now.. so the PR hasn't updated even though I pushed a new commit.
https://status.github.com/

github_fail

marekcirkos:
@facebook-github-bot shipit

PR #135, opened 2016-04-14
Removed buidling modules and added tests to WebDriverAgentLib scheme

marekcirkos:


marekcirkos:
@facebook-github-bot shipit

PR #139, opened 2016-04-21
added elementFromProperty and elementsFromProperty

rafael-chavez:
Example usage:

Lets say we want all the collection views from main window.
First we query for the window we want and get and element id for it using the existing /element endpoint.
we get a response as such:

```
  "value" : [
    {
      "label" : null,
      "type" : "XCUIElementTypeWindow",
      "identifier" : "MyWindow",
      "ELEMENT" : 4
    },
```

We now make a request like such:
```
curl -X POST -H "Content-Type: application/json" -d'{"prop":"collectionViews"}' http://127.0.0.1:8100/session/:sessionId/element/4/elementsFromProperty
```

### DIFF
--- a/WebDriverAgentLib/Commands/FBFindElementCommands.m
+++ b/WebDriverAgentLib/Commands/FBFindElementCommands.m
@@ -91,7 +91,7 @@ static id<FBResponsePayload> FBNoSuchElementErrorResponseForRequest(FBRouteReque
     if (!elements) {
         return FBResponseDictionaryWithStatus(FBCommandStatusNoSuchElement, @{
                                                                               @"description": @"unable to find an elements",
-                                                                              @"sel": request.arguments[@"sel"] ?: @"",
+                                                                              @"property": request.arguments[@"prop"] ?: @"",
                                                                               });
     }
     NSMutableArray *elementsResponse = [[NSMutableArray alloc] init];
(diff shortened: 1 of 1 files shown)


rafael-chavez:
@marekcirkos

rafael-chavez:
@marekcirkos can we ship it?

rafael-chavez:
@marekcirkos can we ship this?

marekcirkos:
I am sorry for delay. I think calling selectors defied by HTTP request directly without much checks (except for respond to selector) is a bit crazy and insecure. In that manner you can call anything e.g. -[XCUIElement resolve] that will probably crash WDA. Also any refactor to elements that we cache will break your endpoint e.g. if we cache XCElementSnapshots or whatever else that is not XCUIElement.

rafael-chavez:
Fair enough @marekcirkos. Do you think it would help if we limit the selectors we can execute to the XCUIElementTypeQueryProvider protocol methods? we could have an array with these methods as strings and do a simple:
```
[arrayWithQueryProviderMethods containsObject:customSelector];
```

marekcirkos:
I think simplest would be something like this https://github.com/facebook/WebDriverAgent/blob/master/WebDriverAgentLib/Categories/XCUIElement%2BWebDriverAttributes.m#L26. However I am more worried about refactor issue. Also keep in mind that this endpoint is basically search by class name which is already supported.

marekcirkos:
I am sorry, but I think it is too risky and does not bring much of benefit

PR #141, opened 2016-04-23
Fix for false status in response from simulator/touch_id

SalmonMode:
When making requests to `simulator/touch-id`, it would return a status of 1 even if the request didn't have any problems. It looks like the reason behind it is because `notify_post(name)` gets evaluated as `False` when the function returns `0` (i.e. `NOTIFY_STATUS_OK`), causing it to jump to the `else` and do that instead. This change resolves that problem by comparing the return value of `notify_post(name)` to   `NOTIFY_STATUS_OK`, instead of just trying to evaluate what the function returns as `True`/`False`.


marekcirkos:
Thank you for PR.

marekcirkos:
@facebook-github-bot shipit

PR #155, opened 2016-05-12
Added endpoint Spec to standard.

rafael-chavez:


rafael-chavez:
@marekcirkos Whenever you get back from holiday, i have reopened pull requests under different name.

WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -181,6 +181,11 @@ + (NSArray *)routes
   NSInteger elementID = [request.parameters[@"id"] integerValue];
   XCUIElement *element = [elementCache elementForIndex:elementID];
   NSError *error = nil;
+  if (!element.fb_isVisible) {
+    if (![element fb_scrollToVisibleWithError:&error]) {
```
marekcirkos: I don't think you expected this here. :)

rafael-chavez: Ugh yea sorry this shouldnt be part of this pull request but now that its come up, do you think it makes sense to scroll to element before doing the "click" command? In the deprecated UIA Framework this was the behavior, also according to the webdriver [spec](https://www.w3.org/TR/webdriver/#dfn-element-click), element click is to scroll to the element into view before clicking.

marekcirkos: Yep, but this implementation will not work, because during scroll element "under" XCUIElement may change. So you might end up tapping wrong element. I suspect that this has something to do with cell reusability. As workaround we don't "autoscroll" when tapping yet and shift scrolling responsibility and identifying correct item to the client. I plan to work on this issue in a wink. I opened [issue](https://github.com/facebook/WebDriverAgent/issues/157) to track it.

rafael-chavez: sounds good i'll remove this and squash my commits

facebook-github-bot-5:
@rafael-chavez updated the pull request.

rafael-chavez:
@marekcirkos updated

rafael-chavez:
Removed unnecessary merge commit.

facebook-github-bot-8:
@rafael-chavez updated the pull request.

marekcirkos:
@facebook-github-bot shipit

facebook-github-bot-4:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1609829742677640/int_phab) to review.

PR #163, opened 2016-05-22
Returning error response, when bundleId is missing in start session command

marekcirkos:


marekcirkos:
@facebook-github-bot shipit

PR #165, opened 2016-06-01
Added DEBUG definition for release build

marekcirkos:
It removes 'undefined macro' warning caused by https://github.com/robbiehanson/KissXML/blob/master/KissXML/DDXML.h#L190


marekcirkos:
@facebook-github-bot import

PR #174, opened 2016-06-09
Kill FBAssertMainThread

marekcirkos:
All requests should be handles on main thread, so checking everywhere for main thread is kind of redundant and pointless as we explicitly dispatch them on on main thread. Much better approach will be writing unit test.


marekcirkos:
@facebook-github-bot import

facebook-github-bot-3:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1813553202191206/int_phab) to review.

PR #202, opened 2016-07-14
Fix inspector's cannot render highlight node on screenshot issue.

chenhengjie123:
In inspector, it may highlight nodes that out of bound like below:

image

It may confuse user since it's not on the screenshot.

After fix, these nodes would never be highlighted.


marekcirkos:
Should we try to fix it rather than hiding? Why it highlights wrong place?

chenhengjie123:
Because the element tree would get not only elements visible on current screen, but also the elements invisible but already initialized in memory. 

On the screenshot above, there are totally three screens on springboard. Element tree would contain all elements on all three screens. 

There are two ways to fix it:

1. Hide all invisible element in element tree. So there is no way to highlight these nodes. We can hide it in WebDriverAgent server by adding "onlyVisible" argument in command, or just hide it using javascript in inspector.
2. Hide nodes highlight by checking its bounds or its **isVisible** attribute. 

May be changing it to hide all invisible elements would be better, since user cannot recognize what element it is while it's invisible on screenshot. If you agree, I can fix it in this way just in few minutes.

marekcirkos:
Ideally we want to do first approach as with iOS 10 we will not report invisible items anymore. Unfortunately we cannot make this change now as we still depend on it. I'll this change then and we can revisit this topic after iOS 10 migration is finished ;)

marekcirkos:
@facebook-github-bot shipit

facebook-github-bot-6:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/308556336144014/int_phab) to review.

PR #205, opened 2016-07-17
Made FBScrollingTests non-device specific

marekcirkos:
Different devices have different screen sizes which means we can't relay on certain elements being visible after single scroll. Better approach would be checking visibility of 'starting' elements before and after scroll, plus check for visibility of any cell.


marekcirkos:
@facebook-github-bot import

facebook-github-bot-4:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/101999440240877/int_phab) to review internal test results.

PR #221, opened 2016-07-29
Fix broken find element tests

marekcirkos:
Fixes tests were broken due to change in Integration app made while landing those tests.


marekcirkos:
@facebook-github-bot import

PR #222, opened 2016-07-29
Stabilize fb_deactivateWithDuration

marekcirkos:
Deactivate application test is randomly failing, as we might return before application becomes active.
This happens, because it returns immediately after tapping application icon on springboard. We can fix this by simply waiting for springboard to disappear before returning control.


marekcirkos:
@facebook-github-bot import

facebook-github-bot-0:
Thanks for importing. If you are an FB employee go to [Phabricator](https://our.intern.facebook.com/intern/opensource/github/pull_request/1059740944079642/int_phab) to review internal test results.

Issue #228, opened 2016-08-02
issue with Scrolling 


yaphatak:
I see there are 2 issues with this particular code 

https://github.com/facebook/WebDriverAgent/blob/master/WebDriverAgentLib/Categories/XCUIElement%2BFBScrolling.m#L83-L85

1. First we should look for either CollectionView or Table  (Preferably CollectionView as I have seen CollectionView is placed in Table) and then should go for ScrollView as ScrollView is superclass of CollectionView and Table.

This causes scroll fail  in scenarios where parent is child of scrollview (say CollectionView or Table)
 
2. How do we handle situation situation where scrollview object here is not visible itself.I think there should be recursive call 
something like this 
   if(![scrollView fb_isVisible]) {
      [scrollView fb_scrollToVisibleWithNormalizedScrollDistance:FBScrollToVisibleNormalizedDistance error:error];
  }

I cant really do this as scrollview is XCElementSnapshot and not XCUIElement object.

We are facing issue because of these as we cant scroll to element whose parent is not visible.

I am willing to fix this. First one is easy fix. I am not sure how can I fix second one unless we implement scrolling logic on XCElementSnapshot object itself. Hope you guys can guide.

marekcirkos:
Fact that CollectionView and TableView inherits from Scrollview doesn't matter in XCTest. Every element has a type and collectionView is type of `XCUIElementTypeCollectionView` not `XCUIElementTypeScrollView`. Only simple scrollview has type of `XCUIElementTypeScrollView`.

Anyway, If you consider view hierarchy with nested scrollviews, there isn't a golden general solution for scrolling.

Best approach would be to scroll in steps. E.g. request scroll to make nested scrollview visible, than scroll to inner element.

yaphatak:
> Best approach would be to scroll in steps. E.g. request scroll to make nested scrollview visible, than scroll to inner element.

Don't you think this will be crazy because you need to know all your parents and then their visibility.  One will end up writing lots of test code just to scroll one target element. Again I am talking from Appium point of view here. Whenever we did scrollintoview or rather click it always brought that element into view. So we dont have to do write that extra code at test side.

I can give  one more shot to implement this nested scroll in WebDriverAgent itself. But will you guys accept that change? or we can keep that as separate endpoint

yaphatak:
Can we do bare minimum like this ?
` const NSUInteger maxScrollCountParent = 5;

  NSUInteger scrollCountParent = 0;
 
[...]

PR #230, opened 2016-08-02
scroll element into view before clicking it

yaphatak:


facebook-github-bot-2:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -176,6 +176,11 @@ + (NSArray *)routes
   NSString *elementUUID = request.parameters[@"uuid"];
   XCUIElement *element = [elementCache elementForUUID:elementUUID];
   NSError *error = nil;
+  
+  //scrollelement to visiable but ignore errros happen during scroll
+  [element fb_scrollToVisibleWithError:&error];
```
marekcirkos: Why ignore error?

yaphatak: I did see that even though scroll was successful it failed with error.  I thought as long as click works fine we dont have to worry about scroll error.

marekcirkos: I think we definitely should check for error and implement it like this:
```
if(![element fb_scrollToVisibleWithError:&error]) {
   return FBResponseWithError(error);
}
```
If it fails on success it sounds like a bug and I'd rather fix it instead of hiding the problem ;).

facebook-github-bot-7:
@yaphatak updated the pull request.

WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -176,6 +176,12 @@ + (NSArray *)routes
   NSString *elementUUID = request.parameters[@"uuid"];
   XCUIElement *element = [elementCache elementForUUID:elementUUID];
   NSError *error = nil;
+  
+  //scroll element to visiable before clicking
```
marekcirkos: Comment is redundant. Please remove it

marekcirkos:
Looks good to me. What about scrolling errors? Did you manage to notice what fails? We should cover it before landing this.

yaphatak:
IF we cant do as we discussed here https://github.com/facebook/WebDriverAgent/issues/228
Then I will do scroll in appium side and then do click. And in that case we dont need this change.

codeskyblue:
@yaphatak But if we donot scroll automaticly, we have to check if element displayed every time before click, this is very time cost.

yaphatak:
@codeskyblue : I know but if we don't do this (https://github.com/facebook/WebDriverAgent/issues/228#issuecomment-237436517) as part of scrolling then there is no point doing this scroll as it wont always guarantee that element scrolled into view. So it doesnt make lot sense to do scrollinto view here. What @marekcirkos suggested that we should always scroll into view as test client code.

marekcirkos:
Also keep in mind that iOS 10 does not report invisible elements (except for cells). It would be much easier to guarantee that WDA reports only visible ones so there would be no need for client to check visibility.

yaphatak:
Thanks @marekcirkos for explanation. Thing is most of child elements (like statictext, button etc) are enclosed in Table or Collection as cell. So these are always reported even though those are not visible. Anyways, as per your suggestion we will do scroll on appum or test client side and I will close this PR. Thanks for all help.

yaphatak:
@marekcirkos : IF you can accept this https://github.com/facebook/WebDriverAgent/pull/246
We can actually merge this PR as well

PR #231, opened 2016-08-02
if element is not specified pick application root as an element

yaphatak:


facebook-github-bot-7:
Thank you for your pull request and welcome to our community. We require contributors to sign our Contributor License Agreement, and we don't seem to have you on file. In order for us to review and merge your code, please sign up at https://code.facebook.com/cla - and if you have received this in error or have any questions, please drop us a line at cla@fb.com. Thanks!

facebook-github-bot-3:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -221,7 +221,11 @@ + (NSArray *)routes
 {
   FBElementCache *elementCache = request.session.elementCache;
   XCUIElement *element = [elementCache elementForUUID:request.parameters[@"uuid"]];
-
+  
+  //if element is not specified then use root element i.e. application
+  if(element == nil) {
```
marekcirkos: What is use case here? Endpoint requires element `:uuid`, so it does make sense to handle invalid uuid as valid use case.

yaphatak: Thanks. I was trying to do something like what you already have  /tap/uuid where you ignore uuid. I didn't notice that scroll is like /uuid/scroll and not scroll/uuid. I am doing this support scroll call from appium which doesnt not take element. So i was just going to call /0/scroll.
Anyways, I can change this to have separate custom endpoint just /scroll. Is that okay with you guys?

facebook-github-bot-9:
@yaphatak updated the pull request.

facebook-github-bot-9:
@yaphatak updated the pull request.

facebook-github-bot-9:
@yaphatak updated the pull request.

WebDriverAgentLib/Commands/FBElementCommands.m
marekcirkos: Yeah `/tap/uuid` is legacy from iOS-driver that we plan to kill. Where are up for adding new end points however be compliant to WD spec https://w3c.github.io/webdriver/webdriver-spec.html

WebDriverAgentLib/Commands/FBCustomCommands.m
```diff
@@ -64,4 +66,25 @@ + (NSArray *)routes
   return FBResponseWithOK();
 }
 
++ (id<FBResponsePayload>)handleScrollCommand:(FBRouteRequest *)request
+{
+    NSString *const direction = request.arguments[@"direction"];
+    if (direction) {
+        if ([direction isEqualToString:@"up"]) {
```
marekcirkos: Maybe we could extract some of this logic to separate method and reuse it here and in element scroll command?

facebook-github-bot-7:
@yaphatak updated the pull request.

WebDriverAgentLib/Categories/XCUIElement+FBScrolling.h
```diff
@@ -34,6 +34,15 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)fb_scrollRight;
 
 /**
+ Scrolls receiver in given direction
+ 
+ @param direction directon in which reciever will be scrolled.
+ @param error If there is an error, upon return contains an NSError object that describes the problem.
+ @return YES if the operation succeeds, otherwise NO.
+ */
+- (BOOL)fb_scrollInDirection:(NSString *)direction error:(NSError **)error;
```
marekcirkos: Adding it to `XCUIElement+FBScrolling.h` doesn't make much sense specially that we have `scrollDown` etc methods. I would keep it in `FBElementCommands.m`.

WebDriverAgentLib/Commands/FBCustomCommands.m
```diff
@@ -32,6 +33,7 @@ + (NSArray *)routes
     [[FBRoute POST:@"/homescreen"].withoutSession respondWithTarget:self action:@selector(handleHomescreenCommand:)],
     [[FBRoute POST:@"/deactivateApp"] respondWithTarget:self action:@selector(handleDeactivateAppCommand:)],
     [[FBRoute POST:@"/timeouts"] respondWithTarget:self action:@selector(handleTimeouts:)],
+    [[FBRoute POST:@"/scroll"] respondWithTarget:self action:@selector(handleScrollCommand:)],
```
marekcirkos: I would keep it in `FBElementCommands.m` to keep scroll commands together.

WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -233,14 +233,9 @@ + (NSArray *)routes
...
-    if ([direction isEqualToString:@"up"]) {
-      [element fb_scrollUp];
-    } else if ([direction isEqualToString:@"down"]) {
-      [element fb_scrollDown];
-    } else if ([direction isEqualToString:@"left"]) {
-      [element fb_scrollLeft];
-    } else if ([direction isEqualToString:@"right"]) {
-      [element fb_scrollRight];
+    NSError *error;
+    if (![request.session.application fb_scrollInDirection:direction error:&error]) {
```
marekcirkos: Probably you meant `element`.

WebDriverAgentTests/IntegrationTests/FBScrollingTests.m
```diff
@@ -50,6 +50,24 @@ - (void)testSimpleScroll
   FBAssertVisibleCell(@"10");
 }
 
+- (void)testScrollInDirection
```
marekcirkos: This is tested already. You should try testing scrolling by using `application` which will cover your case.

marekcirkos:
What is the use case of this endpoint? Why you can't use 'old' endpoint?

yaphatak:
Appium always supported scroll on application. So this will help continue support that. If you see first version of this PR where I proposed if element is nil then we can use application and I was going to call that /0/scroll.

facebook-github-bot-7:
@yaphatak updated the pull request.

marekcirkos:
Why can't you simply query for Application element at the beginning and that use it to scroll app? Having specified for that  seems redundant.

yaphatak:
Yes but that means we need to 2 calls,  one to get an Application element and then do actual scroll. I did this because I saw that there is already similar endpoint with handleTap and handleDrag where you guys use application element directly.
Let me know if you still think it doesn't make sense. I will just close this PR.

marekcirkos:
I don't think that making one extra call for whole test run is too much to pay in order to keep API simpler. 

If you don't agree with me and really really want this endpoint you can easily extend WDA endpoints by implementing extending library that wraps WDALib. WDA automatically adds endpoints from classes that conforms to `FBCommandHandler` protocol so there is no need to change WDA. This makes WDA easy to extend and keep up-to-date.

According handleTap and handleDrag as I mentioned before those are legacy endpoints inherited from iOS-driver and we plan to kill them as soon as me migrate our client. Our main goal is to make WDA WD spec compliant as much as possible.

PR #246, opened 2016-08-12
updating scrolling logic to find scrollable visible Parent

yaphatak:
@marekcirkos : I just want to make sure that its okay to have this logic for scroll. Here we are not doing any scrolling to find parent element but just find parent element who has visible cell snapshot. Lot of code is repetitive and doesn't follow indentation rules. I can fix that later. First I just want to confirm if this approach is okay with you.
In earlier logic I was doing scroll and in fact that was recursive. Here we are just recursively finding parent element with visible snapshot.

### DIFF
--- a/WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
+++ b/WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
@@ -14,6 +14,7 @@
 #import "FBFindElementCommands.h"
 #import "FBRunLoopSpinner.h"
 #import "FBLogger.h"
+#import "FBMacros.h"
 #import "FBXPathCreator.h"
 #import "XCAXClient_iOS.h"
 #import "XCTestDriver.h"
@@ -25,6 +26,7 @@ static NSString *const kXMLIndexPathKey = @"private_indexPath";
 
 inline static BOOL valuesAreEqual(id value1, id value2);
 inline static BOOL isSnapshotTypeAmongstGivenTypes(XCElementSnapshot* snapshot, NSArray<NSNumber *> *types);
+inline static BOOL doesSnapshotHasMoreThanOneVisibleChildrenSnapshots(XCElementSnapshot* snapshot);
 
 @implementation XCElementSnapshot (FBHelpers)
 
@@ -97,11 +99,14 @@ inline static BOOL isSnapshotTypeAmongstGivenTypes(XCElementSnapshot* snapshot, 
   return snapshot;
 }
 
-- (XCElementSnapshot *)fb_parentMatchingOneOfTypes:(NSArray<NSNumber *> *)types
+- (XCElementSnapshot *)fb_findVisibleParentMatchingOneOfTypes:(NSArray<NSNumber *> *)types
 {
   XCElementSnapshot *snapshot = self.parent;
-  while (snapshot && !isSnapshotTypeAmongstGivenTypes(snapshot, types)) {
-      snapshot = snapshot.parent;
+  while (snapshot) {
+    if (isSnapshotTypeAmongstGivenTypes(snapshot, types) && [snapshot isWDVisible] && doesSnapshotHasMoreThanOneVisibleChildrenSnapshots(snapshot)) {
+      break;
+    }
+    snapshot = snapshot.parent;
   }
   return snapshot;
 }
@@ -135,6 +140,20 @@ inline static BOOL isSnapshotTypeAmongstGivenTypes(XCElementSnapshot* snapshot, 
    if([@(snapshot.elementType) isEqual: types[i]] || [types[i] isEqual: @(XCUIElementTypeAny)]){
        return YES;
    }
+  }
+  return NO;
+}
+
+inline static BOOL doesSnapshotHasMoreThanOneVisibleChildrenSnapshots(XCElementSnapshot* snapshot)
+{
+  NSArray<XCElementSnapshot *> *cellSnapshots = [snapshot fb_descendantsMatchingType:XCUIElementTypeCell];
+  if (cellSnapshots.count == 0) {
+    // In some cases XCTest will not report Cell Views. In that case grabbing descendants and trying to figure out scroll directon from them.
+    cellSnapshots = snapshot._allDescendants;
+  }
+  NSArray<XCElementSnapshot *> *visibleCellSnapshots = [cellSnapshots filteredArrayUsingPredicate:[NSPredicate predicateWithFormat:@"%K == YES", FBStringify(XCUIElement, isWDVisible)]];
+  if (visibleCellSnapshots.count > 1) {
+    return YES;
   }
   return NO;
 }
--- a/WebDriverAgentLib/Categories/XCUIElement+FBScrolling.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBScrolling.m
@@ -91,8 +91,9 @@ const CGFloat FBMinimumTouchEventDelay = 0.1f;
   
   if (scrollView == nil) {
     return
-    [[FBErrorBuilder builder]
-      withDescriptionFormat:@"Failed to find scrollable visible parent"];
+    [[[FBErrorBuilder builder]
+      withDescriptionFormat:@"Failed to find scrollable visible parent"]
+     buildError:error];
   }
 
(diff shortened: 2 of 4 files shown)


yaphatak:
I updated this even further. As I noticed that as long as parent is visible its enough. We may run into scenario where visible parent may still don't have at least 2 visible snapshots as your logic requires. (for this to happen each snapshot need to be really big) But I have not ran into such scenario yet. Current change I submitted should cover most of scenario.

WebDriverAgentLib/Categories/XCUIElement+FBScrolling.m
```diff
@@ -279,4 +281,23 @@ - (BOOL)fb_scrollAncestorScrollViewByVectorWithinScrollViewFrame:(CGVector)vecto
...
+                               @(XCUIElementTypeTable),
+                              ];
+    
+  XCElementSnapshot *scrollView = [self fb_parentMatchingOneOfTypes:possibleParents];
+    
+  if (scrollView == nil) {
+    return nil;
+  } else if (![scrollView fb_isVisible]) {
+    return [scrollView fb_findVisibleScrollableView];
+  } else {
```
YatinP: I can move this logic to `fb_parentMatchingOneOfTypes` as well by renaming it to `fb_findVisiableParentMatchingOneOfTypes`

marekcirkos: I think this is a good idea. Also and some tests to make sure it handles invisible elements.

WebDriverAgentLib/Categories/XCUIElement+FBScrolling.m
```diff
@@ -279,4 +281,23 @@ - (BOOL)fb_scrollAncestorScrollViewByVectorWithinScrollViewFrame:(CGVector)vecto
...
+                               @(XCUIElementTypeScrollView),
+                               @(XCUIElementTypeCollectionView),
+                               @(XCUIElementTypeTable),
+                              ];
+    
+  XCElementSnapshot *scrollView = [self fb_parentMatchingOneOfTypes:possibleParents];
+    
+  if (scrollView == nil) {
+    return nil;
+  } else if (![scrollView fb_isVisible]) {
```
marekcirkos: Please use fast returns. `else` is not needed here as you `return` inside `if` anyway. It will simplify the code

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
```diff
@@ -97,11 +99,14 @@ - (XCElementSnapshot *)fb_parentMatchingType:(XCUIElementType)type
...
 }
 
-- (XCElementSnapshot *)fb_parentMatchingOneOfTypes:(NSArray<NSNumber *> *)types
+- (XCElementSnapshot *)fb_findVisibleParentMatchingOneOfTypes:(NSArray<NSNumber *> *)types
 {
   XCElementSnapshot *snapshot = self.parent;
-  while (snapshot && !isSnapshotTypeAmongstGivenTypes(snapshot, types)) {
-      snapshot = snapshot.parent;
+  while (snapshot) {
+    if (isSnapshotTypeAmongstGivenTypes(snapshot, types) && [snapshot isWDVisible] && doesSnapshotHasMoreThanOneVisibleChildrenSnapshots(snapshot)) {
```
yaphatak: I added `doesSnapshotHasMoreThanOneVisibleChildrenSnapshots ` as i ran into issue where it failed because parent element didn't have more than 1 children snapshot. Let me know if you have alternate suggestion for this. I can change this name `fb_findVisibleParentMatchingOneOfTypes` to `fb_findVisibleParentMatchingOneOfTypesAndHasMoreThanOneVisiableChildren`. Also is it worth to send `visibleCellSnapshots` and `cellSnapshots` along with `scrollview` , together as struct so that we don't have to re-calculate.

WebDriverAgentLib/Categories/XCUIElement+FBScrolling.m
```diff
@@ -80,13 +80,20 @@ - (BOOL)fb_scrollToVisibleWithNormalizedScrollDistance:(CGFloat)normalizedScroll
...
-                              ];
+                               ];
     
-  XCElementSnapshot *scrollView = [self.lastSnapshot fb_parentMatchingOneOfTypes:possibleParents];
+  XCElementSnapshot *scrollView = [self.lastSnapshot fb_findVisibleParentMatchingOneOfTypes:possibleParents];
+  
+  if (scrollView == nil) {
+    return
+    [[FBErrorBuilder builder]
+      withDescriptionFormat:@"Failed to find scrollable visible parent"];
```
marekcirkos: You should also call `buildError:`

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
```diff
@@ -138,3 +143,17 @@ inline static BOOL isSnapshotTypeAmongstGivenTypes(XCElementSnapshot* snapshot,
   }
   return NO;
 }
+
+inline static BOOL doesSnapshotHasMoreThanOneVisibleChildrenSnapshots(XCElementSnapshot* snapshot)
+{
+  NSArray<XCElementSnapshot *> *cellSnapshots = [snapshot fb_descendantsMatchingType:XCUIElementTypeCell];
```
marekcirkos: That will not work with simple scrollviews and (table/collection)views with accessibility stripped cells.

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
marekcirkos: First `fb_findVisibleParentMatchingOneOfTypes` suggest nothing about cells. So checking for them here doesn't make sense.
Second this suggest that there is different underlying issue that would be good to understand why it happens.

marekcirkos:
Please check CI, as it fails to build project

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
yaphatak: Thanks @marekcirkos. I know this `fb_findVisibleParentMatchingOneOfTypes` doesnt suggest anything about cell. And thats why I said I will update that in next commit. I just wanted to get your thought on this.

I think underline issue is "just one child is visible and others are not". In this scenario  its failing here https://github.com/facebook/WebDriverAgent/blob/master/WebDriverAgentLib/Categories/XCUIElement%2BFBScrolling.m#L99-L104

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
yaphatak: I know. And thats why following lines where it will just take all available descendants. Actually I copied this from here https://github.com/facebook/WebDriverAgent/blob/master/WebDriverAgentLib/Categories/XCUIElement%2BFBScrolling.m#L92-L97

yaphatak:
@marekcirkos : I also have this question why does `scrollview` objects gets changed to root application object after calling this (it definitely changes to object with `nil` parent. )https://github.com/facebook/WebDriverAgent/blob/master/WebDriverAgentLib/Categories/XCUIElement%2BFBScrolling.m#L92

marekcirkos:
I am not sure what do you mean. Where do they get change

yaphatak:
What I meant is once I run this `NSArray<XCElementSnapshot *> *cellSnapshots = [scrollView fb_descendantsMatchingType:XCUIElementTypeCell];`, I was expecting only to get `cellSnapshots` but what I observed is `scrollView` object also gets modified. Is this expected behavior?

marekcirkos:
@yaphatak are you still working on this?

marekcirkos:
Dropping. If you want to restart please re-open PR and issue.

yaphatak:
@marekcirkos : Sorry, I missed your earlier ping. I was waiting for your reply on https://github.com/facebook/WebDriverAgent/pull/246#r75993755 and https://github.com/facebook/WebDriverAgent/pull/246#r75993756

yaphatak:
I updated here https://github.com/facebook/WebDriverAgent/pull/283

PR #251, opened 2016-08-22
Turn off autocorrection on IntegrationApp textfields

marekcirkos:
FBKeyboardTests tend to fail, because on slower host text gets corrected while typing.


marekcirkos:
@facebook-github-bot import

PR #253, opened 2016-08-22
Moved visibility tests

marekcirkos:
XCTest on iOS 10.x does not report hidden and alpha hidden elements any more, however off-view cells are still reported. In order to keep visibility tests working on both iOS 9 and iOS 10 this diff moves that test to FBScrollingTests.


marekcirkos:
@facebook-github-bot import

facebook-github-bot-1:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3753369) to review internal test results.

PR #255, opened 2016-08-22
Simple scrollview tests

marekcirkos:
After playing a bit with accessibility stripped cells I noticed that way we simulate stripped cells in integration app is bad accessibility configuration rather valid test case. 
This diff replaces them with tests on simple scrollview that simulate this case appropriately.


marekcirkos:
@facebook-github-bot import

facebook-github-bot-8:
Thanks for importing.If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3755023) to review internal test results.

PR #272, opened 2016-09-06
Fix xcodebuild failure due to exit code 65.

marekcirkos:
Travis often fails with exit code 65. After some digging I found out that it is due to fact that xcodebuild's 'test' command counts build time as simulator booting time so it is easy to get timeout on that step.
Easy solution is to explicitly build first and then run test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ branches:
     - master
 
 env:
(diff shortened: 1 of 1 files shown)


marekcirkos:
@facebook-github-bot import

facebook-github-bot-2:
Thanks for importing. If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3823399) to review internal test results.

PR #295, opened 2016-09-21
Send string to picker when setting value

imurchie:
I'm not at all certain that this is the correct way to do this, but it fixes the problem.

When sending text to a PickerWheel, in order to set it, the array value is not found in the possible values and it is rejected. So concatenate the array into a string.

See https://github.com/appium/appium/issues/6873

### DIFF
--- a/WebDriverAgentLib/Commands/FBElementCommands.m
+++ b/WebDriverAgentLib/Commands/FBElementCommands.m
@@ -136,11 +136,14 @@
   if (!value) {
     return FBResponseWithErrorFormat(@"Missing 'value' parameter");
   }
+  NSString *textToType = value;
+  if ([value isKindOfClass:[NSArray class]]) {
+    textToType = [value componentsJoinedByString:@""];
+  }
   if (element.elementType == XCUIElementTypePickerWheel) {
-    [element adjustToPickerWheelValue:value];
+    [element adjustToPickerWheelValue:textToType];
     return FBResponseWithOK();
   }
-  NSString *textToType = [value componentsJoinedByString:@""];
   NSError *error = nil;
   if (![element fb_typeText:textToType error:&error]) {
     return FBResponseWithError(error);


WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -137,7 +137,8 @@ + (NSArray *)routes
     return FBResponseWithErrorFormat(@"Missing 'value' parameter");
   }
   if (element.elementType == XCUIElementTypePickerWheel) {
-    [element adjustToPickerWheelValue:value];
+    NSString *strValue = [value componentsJoinedByString:@""];
```
marekcirkos: Can you check class and depending on class (String or Array) do what is appropriate? So that it can handle both?
After we migrate client I'll remove this check.

imurchie:
It turns out if I use the same function that is used for other types of elements it works. 

```
NSString *textToType = [value componentsJoinedByString:@""];
```

marekcirkos:
I highly doubt that NSString has `componentsJoinedByString:` selector

marekcirkos:
I agree that is should be an array. However our client sends a String right now. This is why, I want you to make it accept both. Otherwise I wouldn't be able to ship it without slow process of migrating our client.

After this lands I can patch your client at some time and then remove this check and handle only array.

imurchie:
Ok! I'm understanding the problem now. Sorry about that.

I've updated to put the check only when it's an `NSArray`. This works on our side.

marekcirkos:
@facebook-github-bot import

facebook-github-bot-5:
Thanks for importing. If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3907080) to review internal test results.

marekcirkos:
@facebook-github-bot shipit

facebook-github-bot-3:
Thanks for importing. If you are an FB employee go to [Phabricator](https://phabricator.intern.facebook.com/D3907080) to review internal test results.

suganeby:
I have the same issue too. Not able to set value to the picker wheel. I tried 'sendKeys' and 'setValue', none of them worked.
It seems to be an issue with XCodeUITest. But not sure when apple would fix the issue because issue seems to be there for over an year. In XCodeUITest, setting value to the picker wheel is achieved by tapping also. So I dont think this issue would get any attention by apple.

Is there any work around from appium side?

Here is the code that I am using
List wheels = driver.findElements(MobileBy.className("XCUIElementTypePickerWheel"))
wheels.get(0).sendKeys("4")

Issue #306, opened 2016-09-28
XPath queries performance is low


mykola-mokhnach:
Currently WD uses DOM-based KissXML framework to perform XPath lookup. And its performance is quite bad in comparison to other solutions based on SAX, for example libxml2. Some complicated XPath queries take about 10 seconds for me and this happens with pretty simply XML tree. Although, I see that the tree itself is built almost instantly, so the problem must be in the lookup method.
Is it possible to replace KissXML with libxml2 to improve the lookup performance?

marekcirkos:
We are not particular attached to KissXML. If you want to put a diff that improves xpath. I am happy to review it.

marekcirkos:
https://github.com/facebook/WebDriverAgent/pull/308

PR #308, opened 2016-09-29
Improve xpath queries performance

mykola-mokhnach:
Current implementation takes up to 20 seconds to handle complicated XPath locators. With these changes the speed is improved to up to four times for some requests (5s vs 20s).
Changes made:
- KissXML framework is replaced with libxml2. This does not affect location performance very much. Only cases, where UI tree is really big, are affected. But now we have less external dependencies since libxml2 is built-in into iPhone SDK
- Made elements filtering procedure more flexible. This change has huge influence to the performance. I did some performance measurements and it seems, that the most time-costly operation here is snapshots comparison, because it is necessary to call 

``` objc
[element resolve]
```

each time we need to get a snapshot. And this was done for every element in the UI. Now we don't do it and try to mach only elements, whose types can potentially match to snapshot types. Unfortunately, this does not make any difference if we are looking for _XCUIElementTypeAny_, but works pretty well (and fast) for other strictly defined element types, like buttons, tables, etc.

The work is based on bug description https://github.com/facebook/WebDriverAgent/issues/306

P. S. Sorry in advance for possible mistakes in the code, I'm not an ObjectiveC developer :)


marekcirkos:
Exiting changes, I'll give it a proper review tomorrow, but from brief look I would move all XML related stuff to separate class/category. Also consider helper methods that you could write tests for :)

mykola-mokhnach:
@marekcirkos I've moved XPath related code into a separate class extension, improved logging and error handling. Plus added some new integration tests. Anything else to improve?

WebDriverAgent.xcodeproj/project.pbxproj
```diff
@@ -1739,6 +1748,7 @@
 					"$(PROJECT_DIR)/Carthage/Build/Mac",
 					"$(PROJECT_DIR)/Carthage/Build/tvOS",
 				);
+				GCC_TREAT_WARNINGS_AS_ERRORS = YES;
```
marekcirkos: Can you revert those?

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.h
```diff
@@ -11,23 +11,16 @@
...
+extern NSString *const XCElementSnapshotInvalidXPathException;
 /**
- Returns an array of descendants matching given xpath query
-
- @param xpathQuery requested xpath query
- @return an array of descendants matching given xpath query
+ The exception happends if any internal error is triggered during XPath matching procedure
  */
-- (NSArray<XCElementSnapshot *> *)fb_descendantsMatchingXPathQuery:(NSString *)xpathQuery;
+extern NSString *const XCElementSnapshotXPathQueryEvaluationException;
```
marekcirkos: You don't need this here.

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.h
```diff
@@ -11,23 +11,16 @@
...
 /**
- Returns an array of descendants matching given type
-
- @param type requested descendant type
- @return an array of descendants matching given type
+ The exception happends if the provided XPath expession cannot be compiled because of a syntax error
  */
-- (NSArray<XCElementSnapshot *> *)fb_descendantsMatchingType:(XCUIElementType)type;
-
+extern NSString *const XCElementSnapshotInvalidXPathException;
```
marekcirkos: You don't need this here.

WebDriverAgentLib/Categories/XCElementSnapshot+FBXPath.m
```diff
@@ -0,0 +1,264 @@
...
+#import "XCUIElement+FBWebDriverAttributes.h"
+
+const static char *_UTF8Encoding = "UTF-8";
+
+static NSString *const kXMLIndexPathKey = @"private_indexPath";
+static NSString *const topNodeIndexPath = @"top";
+NSString *const XCElementSnapshotInvalidXPathException = @"XCElementSnapshotInvalidXPathException";
+NSString *const XCElementSnapshotXPathQueryEvaluationException = @"XCElementSnapshotXPathQueryEvaluationException";
+
+@implementation XCElementSnapshot (FBXPathHelpers)
```
marekcirkos: Methods in this category are way to big. You should create separate class and split them in to smaller ones. If you also make then unaware of XCTest stuff then you can create a lot of unit tests around them and few integration tests which we already have. Otherwise if something breaks it will be scary to touch does.

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
```diff
@@ -9,85 +9,20 @@
...
 #import "XCUIElement+FBWebDriverAttributes.h"
 
-static NSString *const kXMLIndexPathKey = @"private_indexPath";
-
 inline static BOOL valuesAreEqual(id value1, id value2);
 inline static BOOL isSnapshotTypeAmongstGivenTypes(XCElementSnapshot* snapshot, NSArray<NSNumber *> *types);
 
 @implementation XCElementSnapshot (FBHelpers)
 
-- (NSArray<XCElementSnapshot *> *)fb_descendantsMatchingType:(XCUIElementType)type
```
marekcirkos: Assuming you have separate class for explicit XML stuff, I would keep those in and make them use your class. Same as they use KissXML right now.

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -83,18 +85,42 @@ - (void)descendantsWithProperty:(NSString *)property value:(NSString *)value par
   // XPath will try to match elements only class name, so requesting elements by XCUIElementTypeAny will not work. We should use '*' instead.
   xpathQuery = [xpathQuery stringByReplacingOccurrencesOfString:@"XCUIElementTypeAny" withString:@"*"];
   NSArray *matchingSnapshots = [self.lastSnapshot fb_descendantsMatchingXPathQuery:xpathQuery];
-  NSArray *allElements = [[self descendantsMatchingType:XCUIElementTypeAny] allElementsBoundByIndex];
-  NSArray *matchingElements = [self filterElements:allElements matchingSnapshots:matchingSnapshots];
+  if (nil == matchingSnapshots) {
+    return @[];
+  }
+  // Prefiltering elements speeds up search by XPath a lot, because [element resolve] is the most expensive operation here
```
marekcirkos: Nice catch! Add performance test?

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -83,18 +85,42 @@ - (void)descendantsWithProperty:(NSString *)property value:(NSString *)value par
...
+  NSArray *matchingElements = [self filterElements:elementsByType matchingSnapshots:matchingSnapshots];
   return matchingElements;
 }
 
-- (NSArray<XCUIElement *> *)filterElements:(NSArray<XCUIElement *> *)elements matchingSnapshots:(NSArray<XCElementSnapshot *> *)snapshots
+- (NSDictionary<NSNumber *, NSArray<XCElementSnapshot *> *> *)filterPotentialMatchesByType:(NSArray<XCElementSnapshot *> *)snapshots
+{
+  NSMutableSet *matchingTypes = [NSMutableSet set];
+  [snapshots enumerateObjectsUsingBlock:^(XCElementSnapshot *snapshot, NSUInteger snapshotIdx, BOOL *stopSnapshotEnum) {
+    [matchingTypes addObject:@(snapshot.elementType)];
```
marekcirkos: This cries for Swift ;)

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -83,18 +85,42 @@ - (void)descendantsWithProperty:(NSString *)property value:(NSString *)value par
...
+
+- (NSArray<XCUIElement *> *)filterElements:(NSDictionary<NSNumber *, NSArray<XCElementSnapshot *> *> *)elementsMap matchingSnapshots:(NSArray<XCElementSnapshot *> *)snapshots
 {
   NSMutableArray *matchingElements = [NSMutableArray array];
   [snapshots enumerateObjectsUsingBlock:^(XCElementSnapshot *snapshot, NSUInteger snapshotIdx, BOOL *stopSnapshotEnum) {
+    NSArray *elements = elementsMap[@(snapshot.elementType)];
     [elements enumerateObjectsUsingBlock:^(XCUIElement *element, NSUInteger elementIdx, BOOL *stopElementEnum) {
-      [element resolve];
-      if ([element.lastSnapshot _matchesElement:snapshot]) {
+      id lastSnapshot = [element lastSnapshot];
```
marekcirkos: Why `id` here?
Can you add `fb_lastsnapshot` method to `XCUIElement` helper category that does that? We do same thing in few places.

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -83,18 +85,42 @@ - (void)descendantsWithProperty:(NSString *)property value:(NSString *)value par
...
+    return @[];
+  }
+  // Prefiltering elements speeds up search by XPath a lot, because [element resolve] is the most expensive operation here
+  NSDictionary *elementsByType = [self filterPotentialMatchesByType:matchingSnapshots];
+  NSArray *matchingElements = [self filterElements:elementsByType matchingSnapshots:matchingSnapshots];
   return matchingElements;
 }
 
-- (NSArray<XCUIElement *> *)filterElements:(NSArray<XCUIElement *> *)elements matchingSnapshots:(NSArray<XCElementSnapshot *> *)snapshots
+- (NSDictionary<NSNumber *, NSArray<XCElementSnapshot *> *> *)filterPotentialMatchesByType:(NSArray<XCElementSnapshot *> *)snapshots
```
marekcirkos: This will return mapping from type to array of `XCUIElement` not `XCElementSnapshot`. 
Also method name is a bit confusing. You should consider splitting this method into two steps/methods:
1) get snapshot types from `matchingSnapshots`. Key value coding for array gives you that.
2) create typeToDescendantElementMapping:forTypes:

It will be more readable and will give better testability.

marekcirkos:
One more thing. Can you and changes from https://github.com/facebook/WebDriverAgent/pull/285?

WebDriverAgent.xcodeproj/project.pbxproj
```diff
@@ -8,9 +8,11 @@
 
 /* Begin PBXBuildFile section */
 		44757A851D42CF9700ECF35E /* XCUIDeviceRotationTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 44757A831D42CE8300ECF35E /* XCUIDeviceRotationTests.m */; };
+		7122C8621D9FD9A200FE71A3 /* XCElementSnapshot+FBXPath.h in Headers */ = {isa = PBXBuildFile; fileRef = 7122C8601D9FD9A200FE71A3 /* XCElementSnapshot+FBXPath.h */; };
+		7122C8631D9FD9A200FE71A3 /* XCElementSnapshot+FBXPath.m in Sources */ = {isa = PBXBuildFile; fileRef = 7122C8611D9FD9A200FE71A3 /* XCElementSnapshot+FBXPath.m */; };
+		7174AF041D9D39AF008C8AD5 /* libxml2.tbd in Frameworks */ = {isa = PBXBuildFile; fileRef = 7174AF031D9D39AF008C8AD5 /* libxml2.tbd */; };
 		AD35D01A1CF1418E00870A75 /* RoutingHTTPServer.framework in Copy Frameworks */ = {isa = PBXBuildFile; fileRef = AD42DD2B1CF1238500806E5D /* RoutingHTTPServer.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		AD35D01C1CF141A600870A75 /* KissXML.framework in Copy Frameworks */ = {isa = PBXBuildFile; fileRef = AD42DD271CF120B000806E5D /* KissXML.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
```
mykola-mokhnach: fixed

WebDriverAgent.xcodeproj/project.pbxproj
mykola-mokhnach: reverted

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.h
mykola-mokhnach: removed

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
mykola-mokhnach: done. All the stuff is moved FBXPath utility class now

WebDriverAgentLib/Categories/XCElementSnapshot+FBXPath.m
mykola-mokhnach: Refactored and split big methods into a set of smaller ones. 
Though, not sure about how to cover them with unit tests - mocking is needed there for sure. Would be nice to get some examples.

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
mykola-mokhnach: The difference is visible on bigger test data structures. Testing app is too small to see it :(

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
mykola-mokhnach: done

mykola-mokhnach:
@marekcirkos Please check the most recent commit. 
I've also integrated changes from https://github.com/facebook/WebDriverAgent/pull/285 

Not sure about unit testing though. I'd appreciate it if you can create at least one example of such test with mocked objects, so I can use it as a basis for my tests.

marekcirkos:
@mykola-mokhnach  You can remove dependency on `XCElementSnapshot` by passing object that represents it instead and create method on `XCTestSnapshot` that maps it to that representation.  Then you can create them in unit tests cheaply.

mykola-mokhnach:
@marekcirkos I'm thinking about including OCMock framework into the project, which is able to properly do what we need - to proxy XCElementSnapshot instance. This is also what our devs use for unit testing. 
But, yeah, it might take me a bunch of time, since I don't have any experience working with unit testing in ObjC and there are no examples yet. Maybe, several days. Or weeks. Does it work for you?

marekcirkos:
I would NOT go for Mocking in that case. Introducing new data holder class will improve code readability.

mykola-mokhnach:
OK, then I need an example. No idea how to go further without it.

marekcirkos:
Just create a class like `FBXMLElement` that holds whatever you need.

mykola-mokhnach:
Added unit tests %)

mykola-mokhnach:
@marekcirkos Any other things left to do here?

marekcirkos:
@mykola-mokhnach I am sorry, we are super busy right now. I'll check it as soon as I can.

WebDriverAgentLib/Categories/XCUIElement+FBUtilities.m
```diff
@@ -52,5 +52,15 @@ - (BOOL)fb_obstructsElement:(XCUIElement *)element
   return YES;
 }
 
+- (XCElementSnapshot *) fb_lastSnapshot
+{
+  XCElementSnapshot *lastSnapshot = [self lastSnapshot];
+  if (nil == lastSnapshot) {
+    [self resolve];
+    return [self lastSnapshot];
+  } else {
```
marekcirkos: No need for else. I would do fast return here
```
if (self.lastSnapshot) {
  return self.lastSnapshot
}
[self resolve];
return self.lastSnapshot;
```

WebDriverAgentTests/UnitTests/XCTestElementSnapshot.h
```diff
@@ -0,0 +1,14 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <WebDriverAgentLib/XCElementSnapshot.h>
+
+@interface XCTestElementSnapshot : XCElementSnapshot
```
marekcirkos: Definitely not, this name is confusing and you shouldn't really inherit from `XCElementSnapshot`. I would name it `XCElementDouble` that implements `FBElement` protocol and make your method in `FBXPath` accept `id<FBElement>` instead of `XCElementSnapshot`. Then you can create lightweight elements for testing.

WebDriverAgentTests/UnitTests/XCTestElementSnapshot.m
```diff
@@ -0,0 +1,54 @@
...
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import "XCTestElementSnapshot.h"
+
+@implementation XCTestElementSnapshot
+
+- (id)wdValue
```
marekcirkos: Use properties instead and let tests to change them

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -83,18 +85,43 @@ - (void)descendantsWithProperty:(NSString *)property value:(NSString *)value par
...
+  }
+  // Prefiltering elements speeds up search by XPath a lot, because [element resolve] is the most expensive operation here
+  NSSet *matchedTypes = [XCUIElement getSnapshotsTypes:matchingSnapshots];
+  NSDictionary *elementsByType = [self splitElementsByType:matchingSnapshots byTypes:matchedTypes];
+  NSArray *matchingElements = [self filterElements:elementsByType matchingSnapshots:matchingSnapshots];
   return matchingElements;
 }
 
-- (NSArray<XCUIElement *> *)filterElements:(NSArray<XCUIElement *> *)elements matchingSnapshots:(NSArray<XCElementSnapshot *> *)snapshots
++(NSSet *)getSnapshotsTypes:(NSArray<XCElementSnapshot *> *)matchingSnapshots
```
marekcirkos: 1) This should go to `XCElementSnapshot`. 
2) `NSSet<NSNumber *> *`
3) Write tests

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -83,18 +85,43 @@ - (void)descendantsWithProperty:(NSString *)property value:(NSString *)value par
...
+{
+  NSMutableSet *matchingTypes = [NSMutableSet set];
+  [matchingSnapshots enumerateObjectsUsingBlock:^(XCElementSnapshot *snapshot, NSUInteger snapshotIdx, BOOL *stopSnapshotEnum) {
+    [matchingTypes addObject:@(snapshot.elementType)];
+  }];
+  return matchingTypes;
+}
+
+- (NSDictionary<NSNumber *, NSArray<XCUIElement *> *> *)splitElementsByType:(NSArray<XCElementSnapshot *> *)snapshots byTypes:(NSSet *)types
```
marekcirkos: 1) Make it class method on `XCElementSnapshot`
2) Write tests.

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -83,18 +85,43 @@ - (void)descendantsWithProperty:(NSString *)property value:(NSString *)value par
...
 }
 
-- (NSArray<XCUIElement *> *)filterElements:(NSArray<XCUIElement *> *)elements matchingSnapshots:(NSArray<XCElementSnapshot *> *)snapshots
++(NSSet *)getSnapshotsTypes:(NSArray<XCElementSnapshot *> *)matchingSnapshots
+{
+  NSMutableSet *matchingTypes = [NSMutableSet set];
+  [matchingSnapshots enumerateObjectsUsingBlock:^(XCElementSnapshot *snapshot, NSUInteger snapshotIdx, BOOL *stopSnapshotEnum) {
+    [matchingTypes addObject:@(snapshot.elementType)];
+  }];
+  return matchingTypes;
```
marekcirkos: `matchingTypes.copy`

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -83,18 +85,43 @@ - (void)descendantsWithProperty:(NSString *)property value:(NSString *)value par
...
+  NSMutableDictionary *result = [NSMutableDictionary dictionary];
+  [types enumerateObjectsUsingBlock:^(NSNumber *elementTypeAsNumber, BOOL *stopEnum) {
+    XCUIElementType elementType = (XCUIElementType)elementTypeAsNumber.unsignedIntegerValue;
+    NSArray *descendantsOfType = [[self descendantsMatchingType:elementType] allElementsBoundByIndex];
+    result[elementTypeAsNumber] = descendantsOfType;
+  }];
+  return result.copy;
+}
+
+- (NSArray<XCUIElement *> *)filterElements:(NSDictionary<NSNumber *, NSArray<XCUIElement *> *> *)elementsMap matchingSnapshots:(NSArray<XCElementSnapshot *> *)snapshots
```
marekcirkos: Make it class method. If possible, write some tests, but probably it will be painful

WebDriverAgentLib/Categories/XCUIElement+FBUtilities.h
```diff
@@ -31,6 +32,14 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (BOOL)fb_obstructsElement:(XCUIElement *)element;
 
+/**
+ Gets the most recent snapshot of the current element. The element will be 
+ automatically resolved if the snapshot is not available yet
+ 
+ @return The recent snapshot of the element
+ */
+- (XCElementSnapshot *) fb_lastSnapshot;
```
marekcirkos: Remove space

WebDriverAgentLib/Utilities/FBXPath.m
```diff
@@ -0,0 +1,270 @@
...
++ (NSArray<XCElementSnapshot *> *)findMatchesIn:(XCElementSnapshot *)root withXPathQuery:(NSString *)xpathQuery
+{
+  int rc;
+  xmlTextWriterPtr writer;
+  xmlDocPtr doc;
+  
+  writer = xmlNewTextWriterDoc(&doc, 0);
+  if (NULL == writer) {
+    [FBLogger logFmt:@"Failed to invoke libxml2>xmlNewTextWriterDoc for XPath query \"%@\"", xpathQuery];
+    [FBXPath throwException:XCElementSnapshotXPathQueryEvaluationException forQuery:xpathQuery];
```
marekcirkos: Can you add return after each exception throw? I know it doesn't change any thing but makes is easier to read and "scan".

WebDriverAgentLib/Utilities/FBXPath-Private.h
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ Gets xmllib2-compatible XML representation of n XCElementSnapshot instance
+ 
+ @param root the root element to execute XPath query for
+ @param writer the correspondig libxml2 writer object
+ @param withElementStore an empty dictionary to store indexes mapping
+ @return zero if the method has completed successfully
+ */
++(int)getSnapshotAsXML:(XCElementSnapshot *)root withWriter:(xmlTextWriterPtr)writer withElementStore:(NSMutableDictionary *)elementStore;
```
marekcirkos: According to Apple convention "With"/"For" should be used only in first parameter any following should have plain name eg.
`+ (int)matchXMLForSnapshot: write: elementStore:`

Please fix names everywhere.

WebDriverAgentLib/Utilities/FBXPath.m
```diff
@@ -0,0 +1,270 @@
...
+    return rc;
+  }
+  // The current node should be in the store as well
+  elementStore[topNodeIndexPath] = root;
+  rc = xmlTextWriterEndDocument(writer);
+  if (rc < 0) {
+    [FBLogger logFmt:@"Failed to invoke libxml2>xmlXPathNewContext. Error code: %d", rc];
+    return rc;
+  }
+
```
marekcirkos: Kill space

WebDriverAgentLib/Utilities/FBXPath.m
```diff
@@ -0,0 +1,270 @@
...
+    }
+    XCElementSnapshot *element = [elementStore objectForKey:(id)[NSString stringWithCString:(const char*)attrValue encoding:NSUTF8StringEncoding]];
+    if (element) {
+      [matchingSnapshots addObject:element];
+    }
+  }
+  return matchingSnapshots;
+}
+
++(int)getSnapshotAsXML:(XCElementSnapshot *)root withWriter:(xmlTextWriterPtr)writer withElementStore:(NSMutableDictionary *)elementStore
```
marekcirkos: nit: add spaces after `+`, `-`

WebDriverAgentLib/Utilities/FBXPath.m
```diff
@@ -0,0 +1,270 @@
...
+{
+  int rc;
+  
+  rc = xmlTextWriterStartDocument(writer, NULL, _UTF8Encoding, NULL);
```
marekcirkos: `int rc = xmlTextWriterStartDocument(writer, NULL, _UTF8Encoding, NULL);`

WebDriverAgentTests/UnitTests/FBXPathTests.m
```diff
@@ -0,0 +1,79 @@
...
+#import "FBXPath.h"
+#import "FBXPath-Private.h"
+#import "XCTestElementSnapshot.h"
+
+@interface FBXPathTests : XCTestCase
+@end
+
+@implementation FBXPathTests
+
+- (void)testSnapshotXPathPresentation
```
marekcirkos: Awesome!

WebDriverAgentTests/UnitTests/XCTestElementSnapshot.h
mykola-mokhnach: done. Thanks for the idea

WebDriverAgentTests/UnitTests/XCTestElementSnapshot.m
mykola-mokhnach: added properties

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
mykola-mokhnach: done

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
mykola-mokhnach: p 1) is done

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
mykola-mokhnach: made it class method

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
mykola-mokhnach: p. 2) is done
> This should go to XCElementSnapshot 
Any special reason to do it? This utility method is private, short and is not used anywhere except of this implementation.

WebDriverAgentLib/Utilities/FBXPath.m
mykola-mokhnach: done

WebDriverAgentLib/Utilities/FBXPath.m
mykola-mokhnach: done (please let me know if there is anything I missed)

WebDriverAgentLib/Utilities/FBXPath.m
mykola-mokhnach: done

WebDriverAgentLib/Utilities/FBXPath-Private.h
mykola-mokhnach: seems like fixed

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
marekcirkos: It strongly relates to XCElementSnapshots

WebDriverAgentTests/UnitTests/FBXPathTests.m
mykola-mokhnach: thanks, I'm just studying

mykola-mokhnach:
@marekcirkos Seems like done. Shall I merge recent master to this PR since there are conflicts?

jlipps:
@mykola-mokhnach thanks for all the amazing work here! I know our Appium users will be excited for this to land :-)

bearxc:
any update for this fix  ?

morphysk:
any update on this?

WebDriverAgent.xcodeproj/project.pbxproj
```diff
@@ -691,6 +703,8 @@
 		ADBC39951D07840300327304 /* Doubles */ = {
 			isa = PBXGroup;
 			children = (
+				712A0C891DA3E8DE007D02E5 /* XCElementDouble.h */,
```
marekcirkos: Can you reuse `XCUIElementDouble` and extend it instead?

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.h
```diff
@@ -15,17 +15,20 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  Returns an array of descendants matching given type
-
+
```
marekcirkos: Revert

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
```diff
@@ -36,56 +33,12 @@ @implementation XCElementSnapshot (FBHelpers)
...
-  for (NSUInteger i  = 0; i < [children count]; i++) {
-    XCElementSnapshot *childSnapshot = children[i];
-    NSString *newIndexPath = [indexPath stringByAppendingFormat:@",%lu", (unsigned long)i];
-    elementStore[newIndexPath] = childSnapshot;
-    [xmlElement addChild:[self XMLElementFromElement:childSnapshot indexPath:newIndexPath elementStore:elementStore]];
-  }
-  return xmlElement;
+  NSMutableArray<XCElementSnapshot *> *result = [NSMutableArray array];
+  NSArray* matchedSnapshots = [FBXPath findMatchesIn:self xpathQuery:xpathQuery];
+  [matchedSnapshots enumerateObjectsUsingBlock:^(id<FBElement> snapshot, NSUInteger snapshotIdx, BOOL *stopSnapshotEnum) {
```
marekcirkos: What is the point of this enumeration? Why not just return `matchedSnapshots`?

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
```diff
@@ -122,6 +75,15 @@ - (BOOL)fb_framelessFuzzyMatchesElement:(XCElementSnapshot *)snapshot
     valuesAreEqual(self.placeholderValue, snapshot.placeholderValue);
 }
 
++ (NSSet<NSNumber *> *)fb_getUniqueTypes:(NSArray<XCElementSnapshot *> *)matchingSnapshots
+{
+  NSMutableSet *matchingTypes = [NSMutableSet set];
+  [matchingSnapshots enumerateObjectsUsingBlock:^(XCElementSnapshot* snapshot, NSUInteger snapshotIdx, BOOL *stopSnapshotEnum) {
```
marekcirkos: or `[matchingSnapshots valueForKeyPath:@"@distinctUnionOfObjects.elementType"]`

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -83,18 +85,34 @@ - (void)descendantsWithProperty:(NSString *)property value:(NSString *)value par
...
+
++ (NSArray<XCUIElement *> *)filterElements:(NSDictionary<NSNumber *, NSArray<XCUIElement *> *> *)elementsMap matchingSnapshots:(NSArray<XCElementSnapshot *> *)snapshots
 {
   NSMutableArray *matchingElements = [NSMutableArray array];
   [snapshots enumerateObjectsUsingBlock:^(XCElementSnapshot *snapshot, NSUInteger snapshotIdx, BOOL *stopSnapshotEnum) {
+    NSArray *elements = elementsMap[@(snapshot.elementType)];
     [elements enumerateObjectsUsingBlock:^(XCUIElement *element, NSUInteger elementIdx, BOOL *stopElementEnum) {
-      [element resolve];
-      if ([element.lastSnapshot _matchesElement:snapshot]) {
+      if ([[element fb_lastSnapshot] _matchesElement:snapshot]) {
```
marekcirkos: nit: I would keep dot notation

WebDriverAgentLib/Utilities/FBXPath.m
```diff
@@ -0,0 +1,259 @@
...
+    return rc;
+  }
+  
+  rc = [FBXPath recordElementAttributes:writer forElement:root indexPath:indexPath];
+  if (rc < 0) {
+    return rc;
+  }
+  
+  NSArray *children = @[];
+  if ([root respondsToSelector:@selector(children)]) {
```
marekcirkos: Why you need this?

WebDriverAgentLib/Utilities/FBXPath.m
```diff
@@ -0,0 +1,259 @@
...
+      return rc;
+    }
+  }
+  
+  rc = xmlTextWriterEndElement(writer);
+  if (rc < 0) {
+    [FBLogger logFmt:@"Failed to invoke libxml2>xmlTextWriterEndElement. Error code: %d", rc];
+    return rc;
+  }
+
```
marekcirkos: kill space

WebDriverAgentTests/UnitTests/FBElementSnapshotTests.m
```diff
@@ -0,0 +1,41 @@
...
+  el2.elementType = XCUIElementTypeAlert;
+  el2.wdType = @"XCUIElementTypeAlert";
+  [elements addObject:el2];
+  XCElementDouble *el3 = [XCElementDouble new];
+  [elements addObject:el3];
+  
+  XCTAssertNotEqual(el1.elementType, el2.elementType);
+  XCTAssertFalse([el1.wdType isEqualToString: el2.wdType]);
+  XCTAssertEqual(el1.elementType, el3.elementType);
+  XCTAssertTrue([el1.wdType isEqualToString: el3.wdType]);
```
marekcirkos: Those asserts pointless or I am missing something? You are testing `XCElementDouble` here.

WebDriverAgentLib/Utilities/FBXPath.m
mykola-mokhnach: because FBElement protocol does not define children property

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
mykola-mokhnach: any more details on this? it looks like some hidden objc magic for me and I don't usually use anything that I don't understand.

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
mykola-mokhnach: np - changed

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
mykola-mokhnach: Direct typecasting works here. Thanks for the hint.

WebDriverAgent.xcodeproj/project.pbxproj
mykola-mokhnach: peace of cake ;)

WebDriverAgentLib/Utilities/FBXPath.m
mykola-mokhnach: done

mykola-mokhnach:
@marekcirkos All conflicts have been resolved

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -78,23 +80,62 @@ - (void)descendantsWithProperty:(NSString *)property value:(NSString *)value par
...
+  }
+  // Prefiltering elements speeds up search by XPath a lot, because [element resolve] is the most expensive operation here
+  NSSet *matchedTypes = [XCElementSnapshot fb_getUniqueTypes:matchingSnapshots];
+  NSDictionary *elementsByType = [self splitElementsByType:matchingSnapshots byTypes:matchedTypes];
+  NSArray *matchingElements = [XCUIElement filterElements:elementsByType matchingSnapshots:matchingSnapshots];
   return matchingElements;
 }
 
-- (NSArray<XCUIElement *> *)filterElements:(NSArray<XCUIElement *> *)elements matchingSnapshots:(NSArray<XCElementSnapshot *> *)snapshots
+- (NSDictionary<NSNumber *, NSArray<XCUIElement *> *> *)splitElementsByType:(NSArray<XCElementSnapshot *> *)snapshots byTypes:(NSSet *)types
```
marekcirkos: I would really consider moving this method to `XCElementSnapshot+FBHelpers.h` so that you can write tests on it.

WebDriverAgentLib/Commands/FBFindElementCommands.m
```diff
@@ -103,6 +103,10 @@ + (NSArray *)routes
 
 + (XCUIElement *)elementUsing:(NSString *)usingText withValue:(NSString *)value under:(XCUIElement *)element
 {
+  if ([usingText isEqualToString:@"xpath"]) {
```
marekcirkos: Why you need this here? Same check is in line 123.

WebDriverAgentTests/UnitTests/Doubles/XCUIElementDouble.h
```diff
@@ -8,12 +8,26 @@
...
 
 #import <Foundation/Foundation.h>
+#import "FBElement.h"
+#import <XCTest/XCUIElementTypes.h>
 
 @class XCUIApplication;
 
-@interface XCUIElementDouble : NSObject
-@property (nonatomic, strong) XCUIApplication *application;
+@interface XCUIElementDouble : NSObject<FBElement>
```
marekcirkos: Nice

WebDriverAgentTests/UnitTests/FBElementSnapshotTests.m
marekcirkos: But you setup all `XCUIElementDouble` within the test and then check if those values are correct. Instead you should add more checks on `result`, whether types are as expected etc.

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
```diff
@@ -157,6 +105,15 @@ - (XCElementSnapshot *)fb_parentCellSnapshot
     return targetCellSnapshot;
 }
 
++ (NSSet<NSNumber *> *)fb_getUniqueTypes:(NSArray<XCElementSnapshot *> *)matchingSnapshots
```
marekcirkos: We could make this method even more general by using `NSArray<id  <FBElement>> *` and move it to helper class for `FBElements`

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
marekcirkos: Write tests for this method

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
```diff
@@ -36,56 +33,7 @@ @implementation XCElementSnapshot (FBHelpers)
...
-
-  NSArray *children = snapshot.children;
-  for (NSUInteger i  = 0; i < [children count]; i++) {
-    XCElementSnapshot *childSnapshot = children[i];
-    NSString *newIndexPath = [indexPath stringByAppendingFormat:@",%lu", (unsigned long)i];
-    elementStore[newIndexPath] = childSnapshot;
-    [xmlElement addChild:[self XMLElementFromElement:childSnapshot indexPath:newIndexPath elementStore:elementStore]];
-  }
-  return xmlElement;
+  return (NSArray<XCElementSnapshot *> *)[FBXPath findMatchesIn:self xpathQuery:xpathQuery];
```
marekcirkos: Tests for findMatchesIn

WebDriverAgentLib/Utilities/FBXPath.m
```diff
@@ -0,0 +1,266 @@
...
+  xmlTextWriterPtr writer = xmlNewTextWriterDoc(&doc, 0);
+  if (NULL == writer) {
+    [FBLogger logFmt:@"Failed to invoke libxml2>xmlNewTextWriterDoc for XPath query \"%@\"", xpathQuery];
+    [FBXPath throwException:XCElementSnapshotXPathQueryEvaluationException forQuery:xpathQuery];
+    return nil;
+  }
+  NSMutableDictionary *elementStore = [NSMutableDictionary dictionary];
+  int rc = [FBXPath getSnapshotAsXML:root writer:writer elementStore:elementStore];
+  if (rc < 0) {
+    [FBXPath throwException:XCElementSnapshotXPathQueryEvaluationException forQuery:xpathQuery];
```
marekcirkos:
```
xmlFreeTextWriter(writer);
xmlFreeDoc(doc);
```

WebDriverAgentLib/Utilities/FBXPath.m
marekcirkos: This is a pure hack. Instead it might make sense to extend `FBElement` protocol and add `children`, tho probably `XCUIElement` does not implement it.

iquicktest:
@mykola-mokhnach  this is really fantastic.  Looking forward to this cool stuff.

WebDriverAgentLib/Commands/FBFindElementCommands.m
mykola-mokhnach: I've intentionally created separate method to get only the first xpath match, because the current approach resolves multiple snapshots even if this not needed.

Current implementation in master requires the equal amount of time to return a result for

```
driver.findElement(locator)
```

and

```
driver.findElements(locator)
```

In my implementation the first method call could be much faster, because it only resolves the very first matched snapshot and skips everything else, which saves a lot of time (since _resolve_ is the slowest call in the chain).

WebDriverAgentLib/Categories/XCUIElement+FBFind.h
```diff
@@ -39,6 +39,14 @@ NS_ASSUME_NONNULL_BEGIN
 - (NSArray<XCUIElement *> *)fb_descendantsMatchingXPathQuery:(NSString *)xpathQuery;
 
 /**
+ Returns the very first descendant matching given xpath query
+ 
+ @param xpathQuery requested xpath query
+ @return matched descendant or nil if no matches have been found
+ */
+- (XCUIElement *)fb_firstDescendantMatchingXPathQuery:(NSString *)xpathQuery;
```
mykola-mokhnach: add nullable annotation. See the comment below regarding returning an array

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
mykola-mokhnach: done

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
mykola-mokhnach: done

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
mykola-mokhnach: This is nor rewrotten

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
mykola-mokhnach: done

WebDriverAgentLib/Utilities/FBXPath.m
mykola-mokhnach: fixed

WebDriverAgentLib/Commands/FBFindElementCommands.m
marekcirkos: Ok, I see. You could add BOOL parameter to `+ (NSArray *)elementsUsing:(NSString *)usingText withValue:(NSString *)value under:(XCUIElement *)element` something like `canQuitOnFirstMatch` and past it to `fb_descendantsMatchingXPathQuery`. I don't like copy pasting same code around

marekcirkos: Fix that and we will make another code review round :)

mykola-mokhnach:
@marekcirkos how do you like it now?

mykola-mokhnach:
@marekcirkos I constantly test my changes with our Wire application and I have recently noticed, that Apple bug about element visibility detection:

```
XCTRunner[44439]: Enqueue Failure: UI Testing Failure - Failure fetching attributes for element <XCAccessibilityElement: 0x600000448430> Device element: Error Domain=XCTestManagerErrorDomain Code=13 "Error copying attributes -25202" 
```

can be still reproduced in some special cases, even if my previous fix is included. For example, it happens for all table cells if loading animation is visible in one of them. And this can be a serious performance problem while constructing xml for xpath location, since we have read **wdVisible** attribute value now.  
Shall I remove **visible** attribute from source xml, so this Apple bug does not delays xml creation or shall I keep it and hope that Apple fixes the problem in the future?

marekcirkos:
@mykola-mokhnach I would keep it for now. We can improve it in separate PR by e.g. checking if somebody queries isVisible and collect it only then. However first I want to finish this PR, otherwise it will never land

PrivateHeaders/XCTest/XCAXClient_iOS.h
```diff
@@ -8,6 +8,7 @@
 #import <CoreGraphics/CoreGraphics.h>
 
 @class NSMutableDictionary;
+@class XCAccessibilityElement;
```
marekcirkos: nit: I don't think it is needed.

mykola-mokhnach:
@marekcirkos OK, no pb. I've reported this issue to Apple a while ago. Still no any response...  Do they ever care? :(

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -13,21 +13,32 @@
...
+#import "XCUIElement+FBUtilities.h"
 #import "XCUIElement+FBWebDriverAttributes.h"
 
 @implementation XCUIElement (FBFind)
 
 
 #pragma mark - Search by ClassName
 
-- (NSArray<XCUIElement *> *)fb_descendantsMatchingClassName:(NSString *)className
+- (NSArray<XCUIElement *> *)fb_descendantsMatchingClassName:(NSString *)className shouldReturnAfterFirstMatch:(BOOL)shouldReturnAfterFirstMatch
```
marekcirkos: `shouldReturnAfterFirstMatch` doesn't make sense in non-xpath queries as you query all elements anyway. So there is no gain for added complicity.

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -69,51 +80,87 @@ - (void)descendantsWithProperty:(NSString *)property value:(NSString *)value par
...
+}
+
+- (NSArray<XCUIElement *> *)fb_descendantsMatchingXPathQuery:(NSString *)xpathQuery shouldReturnAfterFirstMatch:(BOOL)shouldReturnAfterFirstMatch
+{
+  NSArray *matchingSnapshots = [self getMatchedSnapshotsByXPathQuery:xpathQuery];
+  if (0 == [matchingSnapshots count]) {
+    return @[];
+  }
+  if (shouldReturnAfterFirstMatch) {
+    matchingSnapshots = @[[matchingSnapshots firstObject]];
```
marekcirkos: Now I got confused. You wanted to use `shouldReturnAfterFirstMatch` in order to avoid process ing whole view, but here you process whole view and grab first element anyway?

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
mykola-mokhnach: actually it does, if you check the actual implementation. The actual query object is "lazy" (only calculated upon calling):
scr

Which means that _elementBoundByIndex_ call should not query all the objects, but rather just one of them. I've measured the actual performance there and this call is really faster, but not so much. So I reported another bug to Apple regarding the actual performance of 
_allElementsBoundByIndex_
and
_elementBoundByIndex_
calls, since I expect the second one to be much faster, because we don't have to enumerate all the descendant elements, but rather only one in our case.

PrivateHeaders/XCTest/XCAXClient_iOS.h
marekcirkos: Probably different file is missing it, but don't lose your time looking for it

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
mykola-mokhnach: yes, exactly. xpath evaluation with libxml is super fast and it takes milliseconds to get either one or multiple matches from there. But then we have _filterElements_ call, which enumerates all the found snapshots and matches them to the actual elements, which is super slow. By decreasing count of actual snapshots to one we can skip all the further _snapshots_  array iterations there, which speeds up the search operation significantly.

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
marekcirkos: I don't thin query is resolving elements, but If you measured it, it would be pointless for me to argue with you :).

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -13,21 +13,32 @@
...
 {
   NSMutableArray *result = [NSMutableArray array];
   XCUIElementType type = [FBElementTypeTransformer elementTypeWithTypeName:className];
   if (self.elementType == type || type == XCUIElementTypeAny) {
     [result addObject:self];
   }
-  [result addObjectsFromArray:[[self descendantsMatchingType:type] allElementsBoundByIndex]];
+  XCUIElementQuery *query = [self descendantsMatchingType:type];
+  if (shouldReturnAfterFirstMatch) {
+    if (0 == [result count]) {
```
marekcirkos: If stay with this. Why count 0 is significant here?

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -69,51 +80,87 @@ - (void)descendantsWithProperty:(NSString *)property value:(NSString *)value par
 
 #pragma mark - Search by Predicate String
 
-- (NSArray<XCUIElement *> *)fb_descendantsMatchingPredicate:(NSPredicate *)predicate {
+- (NSArray<XCUIElement *> *)fb_descendantsMatchingPredicate:(NSPredicate *)predicate shouldReturnAfterFirstMatch:(BOOL)shouldReturnAfterFirstMatch
+{
   XCUIElementQuery *query = [[self descendantsMatchingType:XCUIElementTypeAny] matchingPredicate:predicate];
-  NSArray *childElements = [query allElementsBoundByIndex];
-  return childElements;
+  NSMutableArray *result = [NSMutableArray array];
```
marekcirkos: I really don't like repetition. You should extract this code to helper method in that class.

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
marekcirkos: I missed that, makes sense. Thanks for explanation.

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
marekcirkos: You can simplify this logic by making quick return. Meaning checking for `shouldReturnAfterFirstMatch` and `result.count` before you call `descendantsMatchingType` and return if you found something.

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
mykola-mokhnach: Fixed.

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
mykola-mokhnach: I can add a comment there if you want to make it more clear and obvious

marekcirkos:
Something is wrong with this PR. I can't see changed files. Can you rebase?

mykola-mokhnach:
@marekcirkos done

mykola-mokhnach:
@marekcirkos Can you please take a look. Let's finally finish it.
And regarding visibility attribute inclusion - it may slow down the search. I would like to exclude it by default.

limitcracker:
Hi, great work. Is this included already in Appium after the merge? What I have to do to test it?

marekcirkos:
@mykola-mokhnach for such huge delay. You are doing awesome work, but this change is huge and I need time to properly review it to avoid immediate revert. I'll review it tomorrow on conference :)

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.h
```diff
@@ -22,10 +22,13 @@ NS_ASSUME_NONNULL_BEGIN
 - (NSArray<XCElementSnapshot *> *)fb_descendantsMatchingType:(XCUIElementType)type;
 
 /**
- Returns an array of descendants matching given xpath query
-
- @param xpathQuery requested xpath query
- @return an array of descendants matching given xpath query
+ Returns an array of descendants matching given xpath query. This method will always
+ throw an exception if there is an error during XPath evaluation, so the returned array
+ is nevever expected to be equal to nil
```
marekcirkos: nit: never

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -13,21 +13,38 @@
...
 #import "XCElementSnapshot.h"
 #import "XCElementSnapshot+FBHelpers.h"
+#import "XCUIElement+FBUtilities.h"
 #import "XCUIElement+FBWebDriverAttributes.h"
 
 @implementation XCUIElement (FBFind)
 
++ (NSArray<XCUIElement *> *)lookupDescendants:(XCUIElementQuery *)query shouldReturnAfterFirstMatch:(BOOL)shouldReturnAfterFirstMatch
+{
+  if (shouldReturnAfterFirstMatch) {
```
marekcirkos: Please revert this if:
```
if (!shouldReturnAfterFirstMatch) {
  return query.allElementsBoundByIndex;
}
XCUIElement *matchedElement = [query elementBoundByIndex:0];
if (matchedElement && matchedElement.exists) {
  return @[matchedElement];
}
return @[];
```

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -13,21 +13,38 @@
 #import "FBElementTypeTransformer.h"
 #import "XCElementSnapshot.h"
 #import "XCElementSnapshot+FBHelpers.h"
+#import "XCUIElement+FBUtilities.h"
 #import "XCUIElement+FBWebDriverAttributes.h"
 
 @implementation XCUIElement (FBFind)
 
++ (NSArray<XCUIElement *> *)lookupDescendants:(XCUIElementQuery *)query shouldReturnAfterFirstMatch:(BOOL)shouldReturnAfterFirstMatch
```
marekcirkos: This is category so you should add `fb_` prefix. I would also consider different name, maybe
```
+ (NSArray<XCUIElement *> *)fb_extractMatchingElemetnsFromQuery:(XCUIElementQuery *)query shouldReturnOnlyFirstMatch:(BOOL)shouldReturnAfterFirstMatch
```
?

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -69,51 +86,71 @@ - (void)descendantsWithProperty:(NSString *)property value:(NSString *)value par
...
-  NSArray *childElements = [query allElementsBoundByIndex];
-  return childElements;
+  return [self.class lookupDescendants:query shouldReturnAfterFirstMatch:shouldReturnAfterFirstMatch];
 }
 
 
 #pragma mark - Search by xpath
 
-- (NSArray<XCUIElement *> *)fb_descendantsMatchingXPathQuery:(NSString *)xpathQuery
+- (NSArray<XCElementSnapshot *> *)getMatchedSnapshotsByXPathQuery:(NSString *)xpathQuery
```
marekcirkos: Why you need single line method used only once?

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
```diff
@@ -69,51 +86,71 @@ - (void)descendantsWithProperty:(NSString *)property value:(NSString *)value par
...
+  }
+  // Prefiltering elements speeds up search by XPath a lot, because [element resolve] is the most expensive operation here
+  NSSet *byTypes = wdGetUniqueElementsTypes(matchingSnapshots);
+  NSDictionary *categorizedDescendants = [self categorizeDescendants:byTypes];
+  NSArray *matchingElements = [XCUIElement filterElements:categorizedDescendants matchingSnapshots:matchingSnapshots useReversedOrder:[xpathQuery containsString:@"last()"]];
   return matchingElements;
 }
 
-- (NSArray<XCUIElement *> *)filterElements:(NSArray<XCUIElement *> *)elements matchingSnapshots:(NSArray<XCElementSnapshot *> *)snapshots
++ (NSArray<XCUIElement *> *)filterElements:(NSDictionary<NSNumber *, NSArray<XCUIElement *> *> *)elementsMap matchingSnapshots:(NSArray<XCElementSnapshot *> *)snapshots useReversedOrder:(BOOL)useReversedOrder
```
marekcirkos: `fb_` prefix

WebDriverAgentLib/Categories/XCUIElement+FBUtilities.h
marekcirkos: Space still here

WebDriverAgentLib/Categories/XCUIElement+FBUtilities.h
```diff
@@ -31,6 +33,23 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+/**
+ Categorizes all descendants of the current (self) element based on unique types provided in byTypes set
+ 
+ @param byTypes set of XCUIElementType items
+ @return The dictionary, caontaining arrays descendant elements as values categorized by their types as keys.
+ An empty dictionary will be returned in case if byTypes set has no items
+ */
+- (NSDictionary<NSNumber *, NSArray<XCUIElement *> *> *)categorizeDescendants:(NSSet<NSNumber *> *) byTypes;
```
marekcirkos: `fb_` prefix

WebDriverAgentLib/Routing/FBElement.m
```diff
@@ -7,9 +7,21 @@
...
+
 #import "FBElement.h"
+#import "FBElementTypeTransformer.h"
 
 NSString *wdAttributeNameForAttributeName(NSString *name)
 {
   return [NSString stringWithFormat:@"wd%@", name.capitalizedString];
 }
+
+NSSet<NSNumber *> *wdGetUniqueElementsTypes(NSArray<id<FBElement>> *elements)
```
marekcirkos: We should move both `wdAttributeNameForAttributeName` and `wdGetUniqueElementsTypes` to class methods of a new class maybe `FBElementUtils`?

WebDriverAgentLib/Utilities/FBXPath-Private.h
```diff
@@ -0,0 +1,46 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface FBXPath ()
+
+/**
+ Gets xmllib2-compatible XML representation of n XCElementSnapshot instance
+ 
+ @param root the root element to execute XPath query for
+ @param writer the correspondig libxml2 writer object
+ @param withElementStore an empty dictionary to store indexes mapping
```
marekcirkos: Param name do not match cross this file

WebDriverAgentLib/Utilities/FBXPath.m
```diff
@@ -0,0 +1,265 @@
...
+    [FBXPath throwException:XCElementSnapshotInvalidXPathException forQuery:xpathQuery];
+    return nil;
+  }
+  
+  NSArray *matchingSnapshots = [FBXPath collectMatchingSnapshots:queryResult->nodesetval elementStore:elementStore];
+  xmlXPathFreeObject(queryResult);
+  xmlFreeTextWriter(writer);
+  xmlFreeDoc(doc);
+  if (nil == matchingSnapshots) {
+    [FBXPath throwException:XCElementSnapshotXPathQueryEvaluationException forQuery:xpathQuery];
```
marekcirkos: Why not return empty array?

WebDriverAgentLib/Routing/FBElement.h
```diff
@@ -52,6 +52,9 @@ NSString *wdAttributeNameForAttributeName(NSString *name);
 /*! Whether element is an accessibility container (contains children of any depth that are accessible) */
 @property (nonatomic, readonly, getter = isWDAccessibilityContainer) BOOL wdAccessibilityContainer;
 
+/*! Child elements */
+@property(copy) NSArray *children;
```
marekcirkos: This can crash code since `XCUIElement` does not implement it. I would remove it and make `getSnapshotAsXML` accept snapshots. In tests you force cast it from double to Snapshot

WebDriverAgentLib/Utilities/FBXPath.m
mykola-mokhnach: Because this error means there are some matches found by direct xpath matching, but XCTest is unable to match these matches to the actual tree, which is exceptional situation.

WebDriverAgentLib/Routing/FBElement.m
mykola-mokhnach: moved

WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.h
mykola-mokhnach: fixed

WebDriverAgentLib/Categories/XCUIElement+FBFind.m
mykola-mokhnach: fixed

WebDriverAgentLib/Routing/FBElementUtils.h
```diff
@@ -0,0 +1,35 @@
...
+
+@interface FBElementUtils : NSObject
+
+/**
+ Function that returns property name defined by FBElement protocol for given WebDriver Spec property name
+ 
+ @param name WebDriver Spec property name
+ @return the corresponding property name
+ */
++ (NSString *)fb_attributeNameForAttributeName:(NSString *)name;
```
marekcirkos: fb_prefix is not needed here as this is you own class.

WebDriverAgentLib/Utilities/FBXPath.m
marekcirkos: Makes sense :)

marekcirkos:
Looks good. I'll grab your diff and play with it. We should be able to land it next week.

senevit4:
@marekcirkos , hello, any news about landing this pull request?

marekcirkos:
@facebook-github-bot shipit

PR #309, opened 2016-09-30
Updated FBConfiguration.m

sugatmankar:
Instead of fixed default port 8100, allow random ports allocation between 8100 - 8200.
Random port allow to use multiple instance at a time for different devices. 
Randomising port is very useful for iOS real device parallel appium Automation.

### DIFF
--- a/WebDriverAgentLib/Utilities/FBConfiguration.m
+++ b/WebDriverAgentLib/Utilities/FBConfiguration.m
@@ -16,7 +16,7 @@
 
 BOOL _AXSAutomationSetFauxCollectionViewCellsEnabled(BOOL);
 
-static NSUInteger const DefaultStartingPort = 8100;
+//static NSUInteger const DefaultStartingPort = 8100;
 static NSUInteger const DefaultPortRange = 100;
 
 @implementation FBConfiguration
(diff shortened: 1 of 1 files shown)


marekcirkos:
WDA will already bind to first available port between 8100 - 8200. Randomising ports does not improve your situation as you still might have conflicts.

Also you can manage port arrangement on host machine by requesting specific port by 
`USE_PORT` environment variable or `--port` launch argument.

sugatmankar:
Yes it is there but try to hit xcode build command on two different
terminal. Both terminal will start with 8100. This should be avoided.

marekcirkos:
Yes, but you can specify port per terminal

marekcirkos:
Also with your approach every time you call `bindingPortRange` it will give different port, which is not expected.

sugatmankar:
What I was trying to do is "if  I start WDA for one device
```
xcodebuild -project WebDriverAgent.xcodeproj \ -scheme WebDriverAgentRunner
\ -destination 'platform=iOS Simulator,name=iPhone 6' \ test
```
it starts with URL localhost:8100 and for other device it starts with same
url localhost:8100 that's makes appium behavior weird. Appium says WDA is
already running and it won't run new instance.
How to pass port in xcode build command?

PR #310, opened 2016-10-02
scroll element into view before performing any action

yaphatak:


marekcirkos:
This is scary, but I think scroll got much better so we can try.
However I will have to delay it as we are making few changes around WDA and I want to test them before merging this diff.

yaphatak:
1.  I did submit [PR](https://github.com/facebook/WebDriverAgent/pull/230) for this before. But I  closed it as I knew `scroll` logic was not good enough then.

2. If element is visible  then `fb_scrollToVisibleWithError` returns on first line itself. So this should not cause any harm

Also I have question `[XCUIELEMNT resolve]`, does this call do anything other then resolving `snapshotobject` for that particular element?
If `YES` then we should move https://github.com/facebook/WebDriverAgent/blob/master/WebDriverAgentLib/Categories/XCUIElement%2BFBScrolling.m#L86 to `line 90` to make it fast as I observed `resolve` call takes time

marekcirkos:
I definitely want to merge it. But I want to ship scrolling changes part by part. To know what broke it (in case something breaks it). 
We can't reply move resolve, because you want check visibility on most-up-to date element.

marekcirkos:
@facebook-github-bot shipit

PR #327, opened 2016-10-12
Fix build fail on case sensitive filesystems

makhov:
If you enabled case sensitive filesystem on your Mac, you can't build WDA, because filepath is be wrong.
It's very small but very important change


marekcirkos:
@facebook-github-bot import

marekcirkos:
@facebook-github-bot shipit

marekcirkos:
Thank you!

PR #351, opened 2016-11-03
Allow long press and double tap to use coordinates

imurchie:
Both long press and double tap currently require an element, but there are use cases for doing them on a random coordinate. Add this capability.

### DIFF
--- a/WebDriverAgentLib/Commands/FBElementCommands.m
+++ b/WebDriverAgentLib/Commands/FBElementCommands.m
@@ -59,6 +59,8 @@
     [[FBRoute POST:@"/uiaElement/:uuid/scroll"] respondWithTarget:self action:@selector(handleScroll:)],
     [[FBRoute POST:@"/uiaTarget/:uuid/dragfromtoforduration"] respondWithTarget:self action:@selector(handleDrag:)],
     [[FBRoute POST:@"/tap/:uuid"] respondWithTarget:self action:@selector(handleTap:)],
+    [[FBRoute POST:@"/touchAndHold"] respondWithTarget:self action:@selector(handleTouchAndHoldCoordinate:)],
+    [[FBRoute POST:@"/doubleTap"] respondWithTarget:self action:@selector(handleDoubleTapCoordinate:)],
     [[FBRoute POST:@"/keys"] respondWithTarget:self action:@selector(handleKeys:)],
     [[FBRoute GET:@"/window/size"] respondWithTarget:self action:@selector(handleGetWindowSize:)],
     // TODO: This API call should be deprecated and replaced with the one above without the extra :uuid parameter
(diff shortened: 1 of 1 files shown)


WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -198,11 +198,19 @@ + (NSArray *)routes
...
+    NSError *error;
+    if (![element fb_scrollToVisibleWithError:&error]) {
+      return FBResponseWithError(error);
+    }
+    [element doubleTap];
+  } else {
+    FBSession *session = request.session;
+    CGVector point = CGVectorMake((CGFloat)[request.arguments[@"x"] doubleValue], (CGFloat)[request.arguments[@"y"] doubleValue]);
+    XCUICoordinate *appCoordinate = [[XCUICoordinate alloc] initWithElement:session.application normalizedOffset:CGVectorMake(0, 0)];
+    XCUICoordinate *doubleTapCoordinate = [[XCUICoordinate alloc] initWithCoordinate:appCoordinate pointsOffset:point];
```
imurchie: Probably, yes.

imurchie: Updated.

marekcirkos:
Would it make more sense to implement different endpoint? You need to pass uuid for element that does not exist and pass x and y. It doesn't make much sense to me.

marekcirkos:
To be honest, touching at given coordinates is asking for trouble. It is sensible to screen size changes etc.

imurchie:
Yes, it depends on screen, etc. But people want to do it.  ¯\_(ツ)_/¯

As for a different endpoint, I was modeling it after `tap`: https://github.com/facebook/WebDriverAgent/blob/master/WebDriverAgentLib/Commands/FBElementCommands.m#L290-L310

imurchie:
Also something like `/uiaTarget/:uuid/dragfromtoforduration` (https://github.com/facebook/WebDriverAgent/blob/master/WebDriverAgentLib/Commands/FBElementCommands.m#L60) which takes an element but does not use it at all (https://github.com/facebook/WebDriverAgent/blob/master/WebDriverAgentLib/Commands/FBElementCommands.m#L277-L288).

codeskyblue:
+1

marekcirkos:
@imurchie this is legacy endpoint on "to kill" list. I would make separate endpoint, that may or may not share part of implementation.

amedvedjev:
@marekcirkos 
Real usage case. iOS has problems with rendering some elements inside another elements. But we need to tap this element inside another one. The only workaround is find parent element x,y and shift accordingly x or y. Thus dependency from resolution = 0.

imurchie:
Moved to `/touchAndHold` and `/doubleTap` routes, respectively. Thanks!

WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -198,11 +200,16 @@ + (NSArray *)routes
 {
   FBElementCache *elementCache = request.session.elementCache;
```
marekcirkos: Could you keep implementation separate?

imurchie: Certainly! Updated.

PR #358, opened 2016-11-09
Visibility detection fix #2

### LINKED ISSUE #2: Documentation: No need for grep
You if you want to spawn on a booted device you can simply use the `booted` flag, e.g.
```xcrun simctl spawn booted /Users/mehdi/src/WebDriverAgent/Build/...```

No need for grep :wink:

### PR DESCRIPTION
**Pull request by mykola-mokhnach**
Add additional check to workaround Apple's issue about visibility attribute detection.

### CODE REVIEW

File: WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m [hunk 1]
```diff
@@ -16,6 +16,10 @@ @implementation XCUIElement (FBIsVisible)
 
 - (BOOL)fb_isVisible
 {
+  // Calling resolve method is expensive, so we try to perform fast verification first
+  if (!self.exists || CGRectIsEmpty(self.frame) || !CGRectIntersectsRect(self.frame, self.application.frame)) {
```
marekcirkos: Wouldn't calling `exists` resolve the element?

WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
```diff
@@ -28,15 +32,15 @@ @implementation XCElementSnapshot (FBIsVisible)
 
 - (BOOL)fb_isVisible
 {
-  if (self.frame.size.height <= 0 || self.frame.size.width <= 0) {
+  if (CGRectIsEmpty(self.frame) || !CGRectIntersectsRect(self.frame, self.application.frame)) {
```
marekcirkos: Is this optimization or resolving visibility fix (or both)?

mykola-mokhnach: This does not resolve it fully, but just rather makes it to appear less frequently. There are still some cases when this happens and there is no way to prevent these. For example, existence of some active background thread, that changes the UI, while XCTest tries to get accessibility element attributes.

Also, it seems like there are good news about this issue has been fixed in iOS 10.1, but I personally haven't verified it yet. That is why I would advice to disable visibility attribute appending to XML tree (my other PR) by default. Otherwise this might make xpath search even slower than before, even with my fixes :(

Useful links:
https://forums.developer.apple.com/thread/6437
http://stackoverflow.com/questions/36798116/assertion-failure-ui-testing-failure-failure-fetching-attributes-for-element

marekcirkos:
@facebook-github-bot shipit

mykola-mokhnach:
@marekcirkos Please include my recent commit. I've noticed there is another issue, which causes unexpected delays after _self.application.frame_ is invoked. Something like this:
Nov 10 10:48:35 Nick XCTRunner[83698]: Enqueue Failure: UI Testing Failure - Failure getting snapshot Error Domain=XCTestManagerErrorDomain Code=9 "Error getting main window -25204" UserInfo={NSLocalizedDescription=Error getting main window -25204} <unknown> 0 1

That is why I replaced it with 
```
CGRectIsEmpty(self.visibleFrame)
```

which is much more stable.

Sorry again, but it's just physically not possible to review all possible cases. In some UI states everything works just fine, but in some others I start getting such errors.

marekcirkos:
@facebook-github-bot shipit

PR #369, opened 2016-11-14
The third PR to avoid "Error copying attributes -25202" error

mykola-mokhnach:
I have noticed there are still some sporadical timeouts on _wdVisible_ and _wdAccessible_ calls in my application under test even after previously committed fixes have been merged to master. I have spent a lot of time investigating the cause of this problem and it seems like this only happens if one tries to trigger _attributesForElementSnapshot_ method call while there are active UI animations in the application under test. 
_notifyWhenNoAnimationsAreActiveForApplication_ method of _XCAXClient_iOS_ class seems to be the exact thing we need to address the described issue and to avoid this annoying random failure in logs:

```
Enqueue Failure: UI Testing Failure - Failure fetching attributes for element <XCAccessibilityElement: 0x6080004543a0> Device element: Error Domain=XCTestManagerErrorDomain Code=13 "Error copying attributes -25202" UserInfo={NSLocalizedDescription=Error copying attributes -25202} <unknown> 0 1
```


marekcirkos:
This one is tricky. WDA disables waiting for quiescence which includes animations by default. This is mainly, because XCTest is super slow on applications that have some background animations. You can change that during session creation https://github.com/facebook/WebDriverAgent/blob/master/WebDriverAgentLib/Commands/FBSessionCommands.m#L49. I think instead of waiting it would be better to detect if animation is active and then avoid querying instead of blocking.

marekcirkos:
Or even kill how we detect visibility.

mykola-mokhnach:
This is explicit call, so it is not going to slow down anything unless the developer sees a problem and looks for a solution.
Regarding animation detection - I don't think it is possible to implement. I tries to detect it inside the method, which detects visibility, and it still fails, since the XCUIElement instance, for which it is called, is already cached. The trick is that animations presence detection has to be done **before** _findElement_ call. Otherwise _isVisible_ call will fail even if there are no animations currently active.

mykola-mokhnach:
> Or even kill how we detect visibility.

I agree, it's pretty hacky now, but XCTest itself does not provide anything to detect elements visibility properly. People [suggest](http://stackoverflow.com/questions/33247116/testing-if-an-element-is-visible-with-xcode-7-uitest) to use combination of _exists_ and _hittable_ properties (which don't exist for snapshots). Also, they say this method does not work for hidden table cells and night return unpredictable results for some other element types.

marekcirkos:
I just talked to someone on conference that have persistent test case for that issue. Looks like it fail only on fetch for visibility. We might be missing something when we fetch it. Other properties work fine.

mykola-mokhnach:
For me getting wdAccessible property fails with the same error. I suppose this has something to do with _attributesForElementSnapshot_ method call in general.

marekcirkos:
@mykola-mokhnach I'll not merge this diff for now. Let's investigate this issue properly. It looks fixable. We can always go back to this diff if nothing else will work.

mykola-mokhnach:
@marekcirkos I would be happy to know what else can one do to avoid this error in logs while getting visibility attribute. For now I've just added hardcoded 2-3 seconds delays into all places where animations are possible. With visibility option included into xpaths (the other PR) it will be necessary to add even more hardcoded delays :-@

apankov1:
Hi @marekcirkos, we talked about this issue during the conference (it was great to meet you). At this particular moment, I can't provide you a code example when it fails, but I'll be happy to test any changes you make on our app and report results. I'm not sure if this thread is a good place to have this conversation, just wanted to acknowledge that I can be available to test changes you make.

KBhatti:
@apankov1 did you have a solution for this one yet ?

apankov1:
@KBhatti nope

adamlfs:
@mykola-mokhnach, how to build your version from branch wait_for_animation. 
I got 

> Testing failed:
> 	'RoutingHTTPServer/RoutingConnection.h' file not found
> ** TEST FAILED **

after performing building with command:
`xcodebuild -project WebDriverAgent.xcodeproj -scheme WebDriverAgentRunner -destination 'id=XYZ' test`

PR #397, opened 2016-12-05
Attributes validation

mykola-mokhnach:
In this PR I would like to standardise external attribute accessors and hide internal implementation. Also, it should enhance error messages in case of invalid values.

P. S. It looks like I have made a mistake in my previous PR about predicates verification and broke predicates evaluation for attributes like isVisible, since they have getters and I didn't know that KVO requires getter name instead of the actual property name to read the value. Now it is fixed. I'm sorry for that.


WebDriverAgentLib/Categories/NSExpression+FBFormat.m
```diff
@@ -10,37 +10,24 @@
...
 + (instancetype)fb_wdExpressionWithExpression:(NSExpression *)input
 {
   if ([input expressionType] != NSKeyPathExpressionType) {
     return input;
   }
-  NSString *keyPath = [input keyPath];
-  NSString *actualPropName = [FBElementUtils wdAttributeNameForAttributeName:keyPath];
+  NSString *actualPropName = [input keyPath];
+  NSString *suffix = nil;
   if ([actualPropName containsString:@"."]) {
```
marekcirkos: Instead of querying `actualPropName` twice, you could call `[actualPropName rangeOfString:@"."]` and check location for `NSNotFound` value.

WebDriverAgentLib/Routing/FBElementUtils.m
```diff
@@ -12,16 +12,39 @@
 #import "FBElementUtils.h"
 #import "FBElementTypeTransformer.h"
 
+NSString *const FBUnknownAttributeException = @"FBUnknownAttributeException";
 static NSString *const WD_PREFIX = @"wd";
 
 @implementation FBElementUtils
 
 + (NSString *)wdAttributeNameForAttributeName:(NSString *)name
```
marekcirkos: This method is heavily used. And you are recalculating everything every single time.

WebDriverAgentLib/Categories/NSExpression+FBFormat.m
mykola-mokhnach: fixed

WebDriverAgentLib/Routing/FBElementUtils.m
mykola-mokhnach: how do you like it now?

WebDriverAgentLib/Routing/FBElementUtils.h
```diff
@@ -31,11 +39,12 @@ NS_ASSUME_NONNULL_BEGIN
...
 /**
- Returns all properties of FBElement protocol having 'wd' prefix
+ Returns all properties of FBElement protocol having 'wd' prefix with their getters
  
- @return list of matching property names sorted by declaration order
+ @return dictionary of matching property names with their getter names as values if getter method exists
+ for the corresponding property or [NSNull null] if it does not exist
  */
-+ (NSArray<NSString *> *)wdPropertyNames;
++ (NSDictionary<NSString *, id> *)wdProperties;
```
mykola-mokhnach: not really. This dictionary contains all properties having 'wd' prefixes as keys and the corresponding getter names as values or NSNull is there is no getter defined. This getter is returned to make it compatible with KVO lookup, because objc does not understand property names for such purpose if there is a getter method defined for it.

marekcirkos:
I am sorry @mykola-mokhnach, I will not be able to merge this before Christmas. This diff complicates logic a bit and I don't believe it is necessary. However I will not have time to check it properly.

mykola-mokhnach:
Np @marekcirkos 
I wish you Merry Christmas and hope we can finish this in the new year. I'm always happy to learn something new.

marekcirkos:
@mykola-mokhnach Marry Christmas to you too :)

WebDriverAgentLib/Routing/FBElementUtils.m
```diff
@@ -12,16 +12,33 @@
...
+NSString *const FBUnknownAttributeException = @"FBUnknownAttributeException";
 static NSString *const WD_PREFIX = @"wd";
+static NSString *const OBJC_PROP_GETTER_PREFIX = @"G";
+static NSString *const OBJC_PROP_ATTRIBS_SEPARATOR = @",";
 
 @implementation FBElementUtils
 
 + (NSString *)wdAttributeNameForAttributeName:(NSString *)name
 {
-  if ([name hasPrefix:WD_PREFIX]) {
```
marekcirkos: This is heavy method that is called really often. I would strongly consider building up cache for attribute names, because they will never change, right?

WebDriverAgentLib/Routing/FBElementUtils.m
```diff
@@ -33,20 +50,40 @@ + (NSString *)wdAttributeNameForAttributeName:(NSString *)name
...
-    const char *name = property_getName(property);
-    NSString *nsName = [NSString stringWithUTF8String:name];
-    if (nsName && [nsName hasPrefix:WD_PREFIX]) {
-      [result addObject:nsName];
++ (NSDictionary<NSString *, id> *)wdProperties
+{
+  static NSDictionary *propertiesWithGetters;
+  static dispatch_once_t propGettersToken;
+  dispatch_once(&propGettersToken, ^{
+    NSMutableDictionary *result = [NSMutableDictionary new];
```
marekcirkos: If the only reason for this is fb_isVisible etc. I would consider unifying getter names so that we can avoid this.

mykola-mokhnach: I think it is safer to receive getter names from runtime  - we anyway do it only once, which means it should not slow down the tool. What if somebody add a new getter there or rename an existing one? Then the stuff can be easily broken.

WebDriverAgentLib/Routing/FBElementUtils.m
mykola-mokhnach: good idea. Will do

Issue #403, opened 2016-12-07
Broken XPath


marekcirkos:
After new xpath implementation was merged I noticed that some of our queries are failing.
E.G. `.//XCUIElementTypeSheet//XCUIElementTypeButton[@name="Forgot Password?"]` query nukes WDA down.
@mykola-mokhnach Can you have a look?

mykola-mokhnach:
There might be two potential problems in this xpath:
- remove leading dot character
- replace double quotes with single quotes

Let me know which one helps

mykola-mokhnach:
Also, there should be detailed description of the problem in device logs, which libxml2 itself generates if it does not like a query.

marekcirkos:
It looks like it crashes WDA in a hard way. We should be able to handle this query regardless of double quotes or dot. Easies way would be to add unit/integration test with that particular request.

mykola-mokhnach:
that's unfortunate :(
OK, I'll prepare a test tomorrow

mykola-mokhnach:
This was an [easy one](https://github.com/facebook/WebDriverAgent/pull/404/files). Thanks for the hint 👍

PR #404, opened 2016-12-07
Avoid crash for some special XPath queries by explicitly checking whether result nodeSet is empty or not

mykola-mokhnach:
Based on https://github.com/facebook/WebDriverAgent/issues/403


marekcirkos:
Thank you for fast fix!

PR #409, opened 2016-12-09
Support creating a new session without launching an app

qmfrederik:
We launch applications on iOS devices using instruments, but wanted to use the WebDriverAgent to automate certain actions (like dismissing alerts).

We found that a session will, in most cases, use the foremost application, and that launching an application when creating a session was not a hard requirement.

This PR changes the `handleCreateSession` method so that if no `desiredCapabilities` are present and/or no `bundleId` is specified, a new session is created but an application is not launched.

A session created without desired capabilities can still be used to automate apps; given that they were either already active or are launched through another channel.

Let me know if this is something that could be merged upstream.

### DIFF
--- a/WebDriverAgentLib/Commands/FBSessionCommands.m
+++ b/WebDriverAgentLib/Commands/FBSessionCommands.m
@@ -40,21 +40,28 @@
 + (id<FBResponsePayload>)handleCreateSession:(FBRouteRequest *)request
 {
   NSDictionary *requirements = request.arguments[@"desiredCapabilities"];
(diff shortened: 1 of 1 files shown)


marekcirkos:
@qmfrederik Why you even want to create session, without launching application? You can interact with WDA without explicitly creating session. If you call status it will return "blank" session id you can use.

qmfrederik:
@marekcirkos Cool, didn't know about that. Let me see if the blank session works for me.

qmfrederik:
@marekcirkos The default session works fine, closing this PR. Thanks!

PR #410, opened 2016-12-09
Support clicking on a specific alert button

qmfrederik:
Currently, the WebDriverAgent allows you to either accept or dismiss an alert (a leftover from HTML, where JavaScript alerts only defined two options). 

The logic for deciding on which button to click when accepting or dismissing an alert is based on the location of the button - the web driver agent will click on either the first or the last button in the alert.

This PR attempts to make the webdriveragent a bit more flexible, by adding the following routes:

- `GET /alert/buttons`: this route returns the labels (`wdLabel`, same value as returned by `/element/:elementId/text`) of all buttons shown in the current alert
- `POST /alert/click/:label`: this route will try to close the alert by clicking on the button with label `:label`.

All feedback is welcome.

### DIFF
--- a/WebDriverAgentLib/Commands/FBAlertViewCommands.m
+++ b/WebDriverAgentLib/Commands/FBAlertViewCommands.m
@@ -75,7 +75,7 @@
 + (id<FBResponsePayload>)handleGetAlertButtonsCommand:(FBRouteRequest *)request {
   FBSession *session = request.session;
   FBAlert *alert = [FBAlert alertWithApplication:session.application];
-  NSArray *labels = alert.labels;
+  NSArray *labels = alert.buttonLabels;
   
   if (!labels) {
     return FBResponseWithStatus(FBCommandStatusNoAlertPresent, nil);
--- a/WebDriverAgentLib/Commands/FBAlertViewCommands.m
+++ b/WebDriverAgentLib/Commands/FBAlertViewCommands.m
@@ -25,6 +25,8 @@
     [[FBRoute GET:@"/alert/text"] respondWithTarget:self action:@selector(handleAlertTextCommand:)],
     [[FBRoute POST:@"/alert/accept"] respondWithTarget:self action:@selector(handleAlertAcceptCommand:)],
     [[FBRoute POST:@"/alert/dismiss"] respondWithTarget:self action:@selector(handleAlertDismissCommand:)],
+    [[FBRoute GET:@"/alert/buttons"] respondWithTarget:self action:@selector(handleGetAlertButtonsCommand:)],
+    [[FBRoute POST:@"/alert/click/:label"] respondWithTarget:self action:@selector(handleClickAlertButtonCommand:)],
   ];
 }
 
@@ -59,4 +61,26 @@
   return FBResponseWithOK();
 }
 
++ (id<FBResponsePayload>)handleGetAlertButtonsCommand:(FBRouteRequest *)request {
+  FBSession *session = request.session;
+  FBAlert *alert = [FBAlert alertWithApplication:session.application];
+  NSMutableArray *labels = alert.labels;
+  
+  if (!labels) {
+    return FBResponseWithStatus(FBCommandStatusNoAlertPresent, nil);
+  }
+  return FBResponseWithStatus(FBCommandStatusNoError, labels);
+}
+
++ (id<FBResponsePayload>)handleClickAlertButtonCommand:(FBRouteRequest *)request {
+  FBSession *session = request.session;
+  NSString *label = request.parameters[@"label"];
+  
+  FBAlert *alert = [FBAlert alertWithApplication:session.application];
+  
+  if (![alert clickAlertButton:label withError:nil]) {
+    return FBResponseWithStatus(FBCommandStatusNoAlertPresent, nil);
+  }
+  return FBResponseWithOK();
+}
 @end
--- a/WebDriverAgentLib/Commands/FBAlertViewCommands.m
+++ b/WebDriverAgentLib/Commands/FBAlertViewCommands.m
@@ -56,7 +56,7 @@
     if (![alert clickAlertButton:name error:&error]) {
       return FBResponseWithError(error);
     }
-  } else if (![alert acceptWithError:nil]) {
+  } else if (![alert acceptWithError:&error]) {
     return FBResponseWithError(error);
   }
   return FBResponseWithOK();
@@ -76,7 +76,7 @@
     if (![alert clickAlertButton:name error:&error]) {
       return FBResponseWithError(error);
     }
-  } else if (![alert dismissWithError:nil]) {
+  } else if (![alert dismissWithError:&error]) {
     return FBResponseWithError(error);
   }
   return FBResponseWithOK();
@@ -85,11 +85,11 @@
 + (id<FBResponsePayload>)handleGetAlertButtonsCommand:(FBRouteRequest *)request {
   FBSession *session = request.session;
   FBAlert *alert = [FBAlert alertWithApplication:session.application];
-  NSArray *labels = alert.buttonLabels;
-  
-  if (!labels) {
+
+  if (!alert.isPresent) {
     return FBResponseWithStatus(FBCommandStatusNoAlertPresent, nil);
   }
+  NSArray *labels = alert.buttonLabels;
   return FBResponseWithStatus(FBCommandStatusNoError, labels);
 }
 @end
--- a/WebDriverAgentLib/Commands/FBAlertViewCommands.m
+++ b/WebDriverAgentLib/Commands/FBAlertViewCommands.m
@@ -46,13 +46,17 @@
 {
   FBSession *session = request.session;
   NSString *name = request.parameters[@"name"];
+  FBAlert *alert = [FBAlert alertWithApplication:session.application];
+  NSError *error;
 
-  if ([name length] != 0) {
-    if (![[FBAlert alertWithApplication:session.application] clickAlertButton:name error:nil]) {
-      return FBResponseWithStatus(FBCommandStatusNoAlertPresent, nil);
-    }
-  } else if (![[FBAlert alertWithApplication:session.application] acceptWithError:nil]) {
+  if (!alert.isPresent) {
     return FBResponseWithStatus(FBCommandStatusNoAlertPresent, nil);
+  }
+  if ([name length] != 0) {    if (![alert clickAlertButton:name error:&error]) {
+      return FBResponseWithError(error);
+    }
+  } else if (![alert acceptWithError:nil]) {
+    return FBResponseWithError(error);
   }
   return FBResponseWithOK();
 }
@@ -61,13 +65,18 @@
 {
   FBSession *session = request.session;
   NSString *name = request.parameters[@"name"];
-
-  if ([name length] != 0) {
-    if (![[FBAlert alertWithApplication:session.application] clickAlertButton:name error:nil]) {
-      return FBResponseWithStatus(FBCommandStatusNoAlertPresent, nil);
-    }
-  } else if (![[FBAlert alertWithApplication:session.application] dismissWithError:nil]) {
+  FBAlert *alert = [FBAlert alertWithApplication:session.application];
+  NSError *error;
+    
+  if (!alert.isPresent) {
     return FBResponseWithStatus(FBCommandStatusNoAlertPresent, nil);
+  }
+  if ([name length] != 0) {
+    if (![alert clickAlertButton:name error:&error]) {
+      return FBResponseWithError(error);
+    }
+  } else if (![alert dismissWithError:nil]) {
+    return FBResponseWithError(error);
   }
   return FBResponseWithOK();
 }
--- a/WebDriverAgentLib/Commands/FBAlertViewCommands.m
+++ b/WebDriverAgentLib/Commands/FBAlertViewCommands.m
@@ -48,7 +48,7 @@
   NSString *name = request.parameters[@"name"];
 
   if ([name length] != 0) {
-    if (![[FBAlert alertWithApplication:session.application] clickAlertButton:name withError:nil]) {
+    if (![[FBAlert alertWithApplication:session.application] clickAlertButton:name error:nil]) {
       return FBResponseWithStatus(FBCommandStatusNoAlertPresent, nil);
     }
   } else if (![[FBAlert alertWithApplication:session.application] acceptWithError:nil]) {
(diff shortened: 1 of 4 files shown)


jlipps:
@qmfrederik we really shouldn't pollute the json wire protocol api. The appropriate way to add custom routes is behind a vendor flag, like so:

```
GET /wda/alert/buttons
POST /wda/alert/click
```

(I'd also probably have the `label` for the `POST` route be in the JSON params rather than the url)

This is just a recommendation on the API design, I can't comment on whether @marekcirkos likes the idea of this feature in general.

qmfrederik:
@jlipps Sure, we can tweak the JSON API, happy to update the PR if @marekcirkos likes the idea.

Would be interisting to see how other mobile projects do alert handling, and see if we can all agree on a common API for those dreaded alerts ;)

jlipps:
Theoretically all you need is the existing selenium API, and failing that if alert UI is part of the app UI, you can automate it as you would anything else (find elements, click, type text, etc...)

qmfrederik:
Hmm, interesting. Is that how Appium does it today?

What we found is that alerts are sufficiently different warrant a "concept" in their own.

Imagine we would try to handle alerts as normal UI elements - I'm guessing it woukd require a lot of special cases in the handlers for the /element-related routes (checking whether the alert is part of the normal UI or alert UI).

On the client side, I'm thinking you'll see the same level of complexity.

Makes me think of how we tried to handle hybrid - first trying to "embed" it in the native hierarchy only to find that using windows for WebViews was easiers.

Makes me think - what if we were to handle alerts as windows? Would that be easier?

jlipps:
@qmfrederik one of my points is that alerts are already a part of the standard webdriver API as their own entities. They have their own special endpoints already:

* https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidalert_text
* https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessionidaccept_alert
* https://github.com/SeleniumHQ/selenium/wiki/JsonWireProtocol#sessionsessioniddismiss_alert

These allow you to get and set text of an alert/prompt, and accept or dismiss the alert.

Conceptually, is there more you'd want to do with an alert? Almost by definition, standard system alerts can only be interacted with in these ways. Custom alerts can provide more interaction possibilities, but then those should be accessible via the regular UI channels, and are not special system-level modal controls.

qmfrederik:
@jlipps Yes, I know of the alert handling available in the WebDriver protocol, but I believe it's too limited.

For example, the terms "accept" alert and "dismiss" alert really date back to the "OK" and "Cancel" buttons you have on JavaScript alerts in the browser. But if you have an alert on iOS which asks you whether you want to update your app, and you have the options "Yes", "Not Now" and "Never", how would you handle that?

On iOS, alerts are handled through the [`UIAlertController`](https://developer.apple.com/reference/uikit/uialertcontroller) class, and you definitively can define multiple alert actions.

The current implementation only supports two actions (accept and dismiss) and tries to link them to the actual actions using a rather optimistic algorithm (it always chooses the first or the last action, depending on the kind of alert).

Net, on iOS, there are multiple ways for interacting with standard system alerts, and the protocol currently doesn't give you a nice way to interact with them.

I'm still interested in understanding how you would do it - let's see if I can find a demo app we can both use to reason about and make the discussion a bit more tangible?

jlipps:
That sounds good. @lukeis and @shs96c might also be interested in this discussion since it's spec-related.

lukeis:
I have no problem if you add a vendor specific extension, that's what they're there for. This likely won't be added to the spec though.

As for it being necessary at all, I never thought so, I've always been able to just locate the buttons in the dialog and click them (without using the "Alert" interface).

qmfrederik:
Here's an example (taken from the Apple [User Interface Catalog demo]) that shows what this PR tries to achieve :)

Basically, you have an alert which shows _three_ options:
- Choice One
- Choice Two
- Cancel

So; in this case, what should "accept alert" and "dismiss alert" do? And how would you use the Alert API to click on the button that doesn't map to accept nor dismiss?

(https://developer.apple.com/library/content/samplecode/UICatalog/Introduction/Intro.html#//apple_ref/doc/uid/DTS40007710))
screenshot-2016-12-09-22-19-14

lukeis:
>
>
> So; in this case, what should "accept alert" and "dismiss alert" do? And
> how would you use the Alert API to click on the button that doesn't map to
> accept nor dismiss?
>
The Alert API was a construct to HTML documents alert / confirm / prompt
methods. Even though the native context also calls this dialog an Alert,
they are completely different constructs. In a native context the WebDriver
Alerts API no longer makes sense. I recommend not attempting to use it and
using the normal controls that are available to you, namely using
FindElement and locating the button you desire and then clicking it.

marekcirkos:
@jlipps  I am definitely for not polluting json wire protocol API. Most of what we have right now is legacy stuff from iosdriver.

According alerts there are few things you need to remember:
- You can find and accept app alerts using `findElement` & `click`,
- However that will not work with system alerts (Unless something changed on iOS 10).

I think that instead of creating yet another endpoint it would be better to add parameter that would specify button used to "accept", "decline" alert. This way we still conform to json wire protocol API and gain more control. What do you think @jlipps, @qmfrederik

qmfrederik:
@marekcirkos That sounds good to me!

So if I understand it correctly, what you are saying is that the `/alert/accept` and `/alert/dismiss` routes should take a JSON parameter (let's say, `button`), which contains the text of the button on which to click? If so, I can change the PR so that it does that.

That leaves one issue, which is that for system alerts, you'd still need a way to identify all buttons which are available. Would the `GET /alert/buttons` route be suitable for that?

Thanks!

marekcirkos:
@qmfrederik In stead of button I would use something that already exists or WD specific like `name`.

I guess adding buttons to `/alert/text` would be too big modification, so endpoint will make sense, tho use `/wda/alert/buttons` as it is custom.

@jlipps Any thoughts on this?

qmfrederik:
@marekcirkos sounds like a plan, I'll update the PR in the coming days. Thanks!

marekcirkos:
@qmfrederik ping :)

qmfrederik:
@marekcirkos pong :)

In all seriousness, yes, I'll amend the PR as discussed :)

larsjarek:
@qmfrederik I am curious about progress on this one. Sadly it is blocker for my tests.

qmfrederik:
@marekcirkos Sorry it took so long.

I've updated the PR as discussed:
- Moved the method to list all alert buttons to /wda/alert/buttons
- Merged the ability to click on a specific button into the /alert/accept and /alert/dismiss calls. You can specify on which alert you want to click by passing a name property. I've handled the name property in the FBAlertViewCommands class rather than FBAlert; if you want it done differently let me know and I'll update the PR

Let me know what you think

WebDriverAgentLib/FBAlert.h
```diff
@@ -38,6 +38,11 @@ extern NSString *const FBAlertObstructingElementException;
 - (BOOL)isPresent;
 
 /**
+ Gets the labels of the buttons visible in the alert
+ */
+- (nullable NSMutableArray *)labels;
```
marekcirkos: Please do not use Mutable on public interface

WebDriverAgentLib/FBAlert.h
```diff
@@ -59,6 +64,15 @@ extern NSString *const FBAlertObstructingElementException;
 - (BOOL)dismissWithError:(NSError **)error;
 
 /**
+ Clicks on an alert button, if present
+ 
+ @param label The label of the button on which to click.
+ @param error If there is an error, upon return contains an NSError object that describes the problem.
+ @return YES if the operation suceeds, otherwise NO.
+ */
+- (BOOL)clickAlertButton:(NSString *)label withError:(NSError **)error;
```
marekcirkos: drop `with`

WebDriverAgentLib/FBAlert.m
```diff
@@ -106,6 +106,27 @@ - (NSString *)text
   return mText.length > 0 ? mText.copy : [NSNull null];
 }
 
+- (NSMutableArray *) labels
+{
+  NSMutableArray *value = [NSMutableArray array];
+
```
marekcirkos: Remove spaces and empty lines

WebDriverAgentLib/FBAlert.m
```diff
@@ -147,6 +168,29 @@ - (BOOL)dismissWithError:(NSError **)error
...
 }
 
+- (BOOL)clickAlertButton:(NSString *)label withError:(NSError **)error {
+  
+  XCUIElement *alertElement = self.alertElement;
+  NSArray<XCUIElement *> *buttons = [alertElement descendantsMatchingType:XCUIElementTypeButton].allElementsBoundByIndex;
+  XCUIElement *requestedButton;
+  
+  for(XCUIElement *button in buttons) {
+    if([[button wdLabel] isEqualToString:label]){
```
marekcirkos: Probably we want to do case insensitive comparison?

qmfrederik: I'm not sure. I think most of WDA is currently case insensitive (at least when comparing on labels). Also, consider the - admittedly unlikely - events of having two labels with the same text but different casing. How would we handle that?

qmfrederik:
@marekcirkos I added integration tests and updated the PR to account for your feedback. Let me know if you need anything else.

WebDriverAgentLib/Commands/FBAlertViewCommands.m
```diff
@@ -44,7 +45,13 @@ + (NSArray *)routes
 + (id<FBResponsePayload>)handleAlertAcceptCommand:(FBRouteRequest *)request
 {
   FBSession *session = request.session;
-  if (![[FBAlert alertWithApplication:session.application] acceptWithError:nil]) {
+  NSString *name = request.parameters[@"name"];
+
+  if ([name length] != 0) {
+    if (![[FBAlert alertWithApplication:session.application] clickAlertButton:name error:nil]) {
+      return FBResponseWithStatus(FBCommandStatusNoAlertPresent, nil);
```
marekcirkos: That might be misleading right? We should handle error here properly and message whether alert or this particular button is not present

WebDriverAgentLib/FBAlert.h
```diff
@@ -38,6 +38,11 @@ extern NSString *const FBAlertObstructingElementException;
 - (BOOL)isPresent;
 
 /**
+ Gets the labels of the buttons visible in the alert
+ */
+- (nullable NSArray *)labels;
```
marekcirkos: This is misleading, call it buttonLabels instead?

WebDriverAgentLib/FBAlert.m
marekcirkos: The would be extreme edge case. Tho I get the point about matching labels.

WebDriverAgentLib/Commands/FBAlertViewCommands.m
qmfrederik: I get your point, but I'm not sure how to fix it - would you want to respond with FBCommandStatusNoAlertPresent and a more detailed error message, or do you want to return a different status code - something along the lines of Element not found?

marekcirkos:
I am happy to land it, but you have some build issues on Travis (conversions)

WebDriverAgentLib/Commands/FBAlertViewCommands.m
marekcirkos: I would suggest:
1) Checking if alert is present, and raise `FBCommandStatusNoAlertPresent` if not,
2) Dismiss/Accept with/without button specified and raise `FBResponseWithErrorFormat` in case of failure.

WebDriverAgentLib/FBAlert.m
```diff
@@ -106,6 +106,20 @@ - (NSString *)text
   return mText.length > 0 ? mText.copy : [NSNull null];
 }
 
+- (NSArray *) buttonLabels
```
marekcirkos: kill space after `)`

qmfrederik:
@marekcirkos Can you have another look? I think I addressed your feedback: fixed a whitespace issue, the tests, and made the return value more explicit.

Let me know what you think!

WebDriverAgentLib/Commands/FBAlertViewCommands.m
```diff
@@ -44,19 +45,50 @@ + (NSArray *)routes
...
   FBSession *session = request.session;
-  if (![[FBAlert alertWithApplication:session.application] acceptWithError:nil]) {
+  NSString *name = request.parameters[@"name"];
+  FBAlert *alert = [FBAlert alertWithApplication:session.application];
+  NSError *error;
+
+  if (!alert.isPresent) {
     return FBResponseWithStatus(FBCommandStatusNoAlertPresent, nil);
   }
+  if ([name length] != 0) {    if (![alert clickAlertButton:name error:&error]) {
```
marekcirkos: Please make them to separate lines :)

qmfrederik: Done!

WebDriverAgentLib/Commands/FBAlertViewCommands.m
```diff
@@ -44,19 +45,51 @@ + (NSArray *)routes
...
+  NSError *error;
+
+  if (!alert.isPresent) {
     return FBResponseWithStatus(FBCommandStatusNoAlertPresent, nil);
   }
+  if ([name length] != 0) {
+    if (![alert clickAlertButton:name error:&error]) {
+      return FBResponseWithError(error);
+    }
+  } else if (![alert acceptWithError:nil]) {
```
marekcirkos: Why not pass error here?

WebDriverAgentLib/Commands/FBAlertViewCommands.m
```diff
@@ -44,19 +45,51 @@ + (NSArray *)routes
...
+  NSError *error;
+    
+  if (!alert.isPresent) {
     return FBResponseWithStatus(FBCommandStatusNoAlertPresent, nil);
   }
+  if ([name length] != 0) {
+    if (![alert clickAlertButton:name error:&error]) {
+      return FBResponseWithError(error);
+    }
+  } else if (![alert dismissWithError:nil]) {
```
marekcirkos: Same here

WebDriverAgentLib/Commands/FBAlertViewCommands.m
```diff
@@ -44,19 +45,51 @@ + (NSArray *)routes
...
+  }
   return FBResponseWithOK();
 }
 
++ (id<FBResponsePayload>)handleGetAlertButtonsCommand:(FBRouteRequest *)request {
+  FBSession *session = request.session;
+  FBAlert *alert = [FBAlert alertWithApplication:session.application];
+  NSArray *labels = alert.buttonLabels;
+  
+  if (!labels) {
```
marekcirkos: I would check of alert presence explicitly here. We would have same logic everywhere

qmfrederik:
@marekcirkos I've hopefully addressed the feedback :)

marekcirkos:
Let's land this bad boy :)

Issue #362, opened 2016-11-09
tap coordinate still wrong on landscape mode, iOS 10


codeskyblue:
- iOS: 10.0.1
- Lastest WDA code
- Model: iPhone 6 Plus (MGAJ2CH/A)

Use `iproxy` to communicate to phone.

Should tap `关于本机`, Actually tap `CarPlay`

![image](https://testerhome.com/photo/2016/eb47f80e27717cb5cefd5098f699dd12.png)

marekcirkos:
I broke rotations, so that might be related. However can you post what are frames for both?

marekcirkos:
Found an issue, fix is on the way.

marekcirkos:
https://github.com/facebook/WebDriverAgent/commit/464e7adb32ac3e2cdc24129d4cb4e2f2ec87c834

marekcirkos:
@codeskyblue Please confirm

codeskyblue:
last night I update my iPhone to 10.1.1
So xcode should also update, It may take 6~8 hours because of the slow network.
Sorry I can't confirm recently.

codeskyblue:
last night I update my iPhone to 10.1.1
So xcode should also update, It may take 6~8 hours because of the slow network.
Sorry I can't confirm recently.

codeskyblue:
last night I update my iPhone to 10.1.1
So xcode should also update, It may take 6~8 hours because of the slow network.
Sorry I can't confirm recently.

codeskyblue:
last night I update my iPhone to 10.1.1
So xcode should also update, It may take 6~8 hours because of the slow network.
Sorry I can't confirm recently.

codeskyblue:
Confirm. still the wrong coordinate.

marekcirkos:
Which endpoint are you using?

marekcirkos:
@codeskyblue ?

codeskyblue:
Sorry for my delay, I can not upload the image, because the bad network.

This is my request

```
curl -X POST -d '{"using": "link text", "value": "name=\u652f\u4ed8\u5b9d"}' 'http://127.0.0.1:8100/session/E0C2F178-039B-4585-9FF1-CAC68E8B41AA/elements'
Return: {
    "status": 0, 
    "sessionId": "E0C2F178-039B-4585-9FF1-CAC68E8B41AA", 
    "value": [
        {
            "label": "\u652f\u4ed8\u5b9d", 
[...]

PR #411, opened 2016-12-11
Support different gestures in landscape mode

mykola-mokhnach:
Fixes https://github.com/facebook/WebDriverAgent/issues/362


WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -58,6 +59,8 @@ + (NSArray *)routes
     [[FBRoute POST:@"/uiaElement/:uuid/twoFingerTap"] respondWithTarget:self action:@selector(handleTwoFingerTap:)],
     [[FBRoute POST:@"/uiaElement/:uuid/touchAndHold"] respondWithTarget:self action:@selector(handleTouchAndHold:)],
     [[FBRoute POST:@"/uiaElement/:uuid/scroll"] respondWithTarget:self action:@selector(handleScroll:)],
+    [[FBRoute POST:@"/dragfromtoforduration"] respondWithTarget:self action:@selector(handleDrag:)],
```
marekcirkos: Can you not use different endpoint something like `/wda/dragfromtoforduration`

WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -58,6 +59,8 @@ + (NSArray *)routes
...
+    // TODO: remove obsolete endpoint
     [[FBRoute POST:@"/uiaTarget/:uuid/dragfromtoforduration"] respondWithTarget:self action:@selector(handleDrag:)],
```
marekcirkos: We should split those handler. I don't like that we try to handle both cases in one handle and being smart about element presence.

WebDriverAgentLib/Utilities/FBMathUtils.m
```diff
@@ -52,3 +52,13 @@ CGPoint FBInvertPointForApplication(CGPoint point, CGSize screenSize, UIInterfac
       return CGPointMake(screenSize.width - point.y, point.x);
   }
 }
+
+CGSize FBAdjustDimensionsForApplication(CGSize actualSize, UIInterfaceOrientation orientation)
+{
+  if (orientation == UIInterfaceOrientationLandscapeLeft || orientation == UIInterfaceOrientationLandscapeRight) {
+    if (actualSize.height > actualSize.width) {
```
marekcirkos: ?

WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -350,13 +354,12 @@ + (NSArray *)routes
...
-+ (XCUICoordinate *)getGestureCoordinate:(FBRouteRequest *)request
++ (XCUICoordinate *)gestureCoordinateWithCoordinate:(CGPoint)coordinate application:(XCUIApplication *)application
 {
-  FBSession *session = request.session;
-  CGVector point = CGVectorMake((CGFloat)[request.arguments[@"x"] doubleValue], (CGFloat)[request.arguments[@"y"] doubleValue]);
-  XCUICoordinate *appCoordinate = [[XCUICoordinate alloc] initWithElement:session.application normalizedOffset:CGVectorMake(0, 0)];
-  XCUICoordinate *gestureCoordinate = [[XCUICoordinate alloc] initWithCoordinate:appCoordinate pointsOffset:point];
-  return gestureCoordinate;
+  // TODO: handle this properly after XCTest starts to respect landscape interface orientation
+  CGPoint point = FBInvertPointForApplication(coordinate, application.frame.size, application.interfaceOrientation);
```
marekcirkos: 1) You should check for iOS version here as well.
2) Instead of forcing those checks everywhere, you might want to consider adding `FBUICoordinate` class and patch coordinates on the go?

WebDriverAgentLib/Utilities/FBMathUtils.m
mykola-mokhnach: XCTest bug is, that it returns exchanged dimensions for landscape mode. This verification is just to make sure the bug is still there (since height is never greater than width in landscape) and to make it still working properly after XCTest itself starts to respect landscape.

WebDriverAgentLib/Commands/FBElementCommands.m
mykola-mokhnach: > You should check for iOS version here as well. 
This might be a bad idea, because you never know whether  given coordinates were calculated based on some element position or just based on the actual screen size. I would prefer to keep it like this, because this is generic method and only keep this verification for methods, where we explicitly calculating offset based on element location (like tap)

WebDriverAgentLib/Commands/FBElementCommands.m
mykola-mokhnach: no pb - I will implement both methods

WebDriverAgentLib/Commands/FBElementCommands.m
mykola-mokhnach: I have just created it based on existing API methods examples, like "/touchAndHold" and "/doubleTap". Shouldn't we follow the same template there?

marekcirkos: Those are legacy stuff, that should die.

WebDriverAgentLib/Commands/FBElementCommands.m
marekcirkos: Yes, but all versions checks etc, would happen in that class. So final product would be "transparent" for all clients. All of them should expect correct values anyway. Only problem would be argument you made in first point.

WebDriverAgentLib/Utilities/FBMathUtils.m
marekcirkos: Can you and comment on that?

WebDriverAgentLib/Commands/FBElementCommands.m
mykola-mokhnach: ok then I will use wda/ prefix as you suggested

WebDriverAgentLib/Commands/FBElementCommands.m
mykola-mokhnach: Let's make it simple. Most of users don't even use 20% of features we have here. I could think about improving this if there are enough complains in Appium forum. But for now it might take a lot of time to just create a workaround for the stuff, which is not widely used. And time is money 💯

WebDriverAgentLib/Utilities/FBMathUtils.m
mykola-mokhnach: sure

WebDriverAgentLib/Commands/FBElementCommands.m
mykola-mokhnach: I've also noticed one more thing, that is not going to be resolved if we just transform element coordinates. All the methods of XCUIElement, like _doubleTap_, _pinch_, etc will still fail in landscape. We anyway need to replace them with something, that does the same action by coordinates (actually, this is what you've already done for _click_) :-(

marekcirkos:
I am sorry @mykola-mokhnach, I will not be able to merge this before Christmas. There are few things I would like do differently, but I wouldn't have time to sit to properly.

codeskyblue:
@marekcirkos Any progress on this pr?

WebDriverAgentLib/Categories/XCUIElement+FBTap.h
```diff
@@ -21,6 +21,15 @@ NS_ASSUME_NONNULL_BEGIN
...
 - (BOOL)fb_tapWithError:(NSError **)error;
 
+/**
+ Waits for element to become stable (not move) and performs sync tap on element
+ 
+ @param error If there is an error, upon return contains an NSError object that describes the problem.
+ @param relativeCoordinate hit point coordinate relative to the current element position
+ @return YES if the operation succeeds, otherwise NO.
+ */
+- (BOOL)fb_tapCoordinateWithError:(NSError **)error relativeCoordinate:(CGPoint)relativeCoordinate;
```
marekcirkos: Error parameter should be last.
```
- (BOOL)fb_tapAtRelativeCoordinate:(CGPoint)relativeCoordinate error:(NSError **)error;
```

WebDriverAgentLib/Categories/XCUIElement+FBTap.m
```diff
@@ -21,12 +22,23 @@ @implementation XCUIElement (FBTap)
...
+- (BOOL)fb_tapCoordinateWithError:(NSError **)error relativeCoordinate:(CGPoint)relativeCoordinate
+{
   [self fb_waitUntilFrameIsStable];
   __block BOOL didSucceed;
   [FBRunLoopSpinner spinUntilCompletion:^(void(^completion)()){
-    NSValue *hitpointValue = self.lastSnapshot.suggestedHitpoints.firstObject;
-    CGPoint hitPoint = hitpointValue ? hitpointValue.CGPointValue : [self coordinateWithNormalizedOffset:CGVectorMake(0.5, 0.5)].screenPoint;
-    hitPoint = FBInvertPointForApplication(hitPoint, self.application.frame.size, self.application.interfaceOrientation);
+    CGPoint hitPoint = CGPointMake(self.frame.origin.x + relativeCoordinate.x, self.frame.origin.y + relativeCoordinate.y);
+    // TODO: Change that after XCTest starts to respect landscape orientation
```
marekcirkos: Instead of todo I would add comment why this is needed.

WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -206,7 +208,8 @@ + (NSArray *)routes
 
 + (id<FBResponsePayload>)handleDoubleTapCoordinate:(FBRouteRequest *)request
 {
-  XCUICoordinate *doubleTapCoordinate = [self.class getGestureCoordinate:request];
+  CGPoint doubleTapPoint = CGPointMake((CGFloat)[request.arguments[@"x"] doubleValue], (CGFloat)[request.arguments[@"y"] doubleValue]);
+  XCUICoordinate *doubleTapCoordinate = [self.class gestureCoordinateWithCoordinate:doubleTapPoint application:request.session.application shouldApplyOrientationWorkaround:YES];
```
marekcirkos: Why you always apply changes here without iOS check? Would it be better to make iOS check inside that method?

WebDriverAgentLib/Categories/XCUIElement+FBTap.h
mykola-mokhnach: fixed

WebDriverAgentLib/Categories/XCUIElement+FBTap.m
mykola-mokhnach: added detailed comment

WebDriverAgentLib/Commands/FBElementCommands.m
mykola-mokhnach: Because of the same issue that I've already described. We never know how these coordinates were calculated by the client - based on the absolute screen width/height or based on the element position. But nevermind - I will add this check since we have included a fix for screen size into this PR.

mykola-mokhnach: Sorry, please ignore my recent comment. This is done, because in iOS 9.3 drag coordinates are also inverted. So the current implementation works properly if one provides absolute coordinates based on the actual screen width/height, but will not work if relative coordinates based on some element position are provided. I think the first case is more relevant for such method call. If one wanted to have a method that clicks some element by relative coordinates, then he would create some other method (something like _tap_). But this method only accepts absolute coordinates

@imurchie waht is you opinion on it?

marekcirkos: Just to summarise what we know:
1) On iOS 9.x coordinates are not inverted
2) On iOS 10.x coordinates are inverted
3) We can assume that coordinates passed by user are correct (meaning they are in reference to current orientation). This is mainly because we give them the coordinates or they manually calculate them.
Is that right?

mykola-mokhnach: > 1. On iOS 9.x coordinates are not inverted

Yes, but only coordinates we get when we want to know XCUIElement frame. All the other XCUIElement methods have the same bug as in 10.0+

> On iOS 10.x coordinates are inverted

Yes, but this might be fixed in any of upcoming iOS releases.

> 3. We can assume that coordinates passed by user are correct (meaning they are in reference to current orientation). This is mainly because we give them the coordinates or they manually calculate them.

I want to discuss this point with @imurchie, because most of WDA calls where we do operations with absolute screen coordinates cannot be invoked directly from Appium, but rather via TouchAction interface. And we need to be in sync about our expectations here.

marekcirkos:
@mykola-mokhnach is it ready for review or we are waiting for @imurchie?

mykola-mokhnach:
Yes, this is ready from my side. I just want @imurchie to take a final look at this PR, because he was working on TouchAction interface in Appium, which, actually, is the main consumer of fixed WDA API methods.

imurchie:
Looks good to me! This will be awesome to get in!

mykola-mokhnach:
Hi @marekcirkos 
Can you please initiate marge procedure for this PR? I'd like to create another PR for visibility fi, but it uses some stuff from this one.

WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -275,34 +279,47 @@ + (NSArray *)routes
...
-  CGVector endPoint = CGVectorMake((CGFloat)[request.arguments[@"toX"] doubleValue], (CGFloat)[request.arguments[@"toY"] doubleValue]);
+  FBElementCache *elementCache = session.elementCache;
+  XCUIElement *element = [elementCache elementForUUID:request.parameters[@"uuid"]];
+  CGPoint startPoint = CGPointMake((CGFloat)(element.frame.origin.x + [request.arguments[@"fromX"] doubleValue]), (CGFloat)(element.frame.origin.y + [request.arguments[@"fromY"] doubleValue]));
+  CGPoint endPoint = CGPointMake((CGFloat)(element.frame.origin.x + [request.arguments[@"toX"] doubleValue]), (CGFloat)(element.frame.origin.y + [request.arguments[@"toY"] doubleValue]));
   NSTimeInterval duration = [request.arguments[@"duration"] doubleValue];
-  XCUICoordinate *appCoordinate = [[XCUICoordinate alloc] initWithElement:session.application normalizedOffset:CGVectorMake(0, 0)];
-  XCUICoordinate *endCoordinate = [[XCUICoordinate alloc] initWithCoordinate:appCoordinate pointsOffset:endPoint];
-  XCUICoordinate *startCoordinate = [[XCUICoordinate alloc] initWithCoordinate:appCoordinate pointsOffset:startPoint];
+  BOOL shouldApplyOrientationWorkaround = SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(@"10.0");
```
marekcirkos: Why we care about iOS version if we are handling coordinates there that are always buggy?

mykola-mokhnach: here we calculate coordinates based on _element_'s position, which is calculated properly for iOS 9.3 and is broken for iOS10.0+ 
In other cases coordinates are passed as absolute values.

marekcirkos: Ok, let's do this then!

PR #426, opened 2016-12-22
Add device ip to status record

imurchie:
It is useful to know what IP the device has, without needing to parse the log file for the server's startup message.

I'm not sure the best place within the returned data to put this.

### DIFF
--- a/WebDriverAgentLib/Commands/FBSessionCommands.m
+++ b/WebDriverAgentLib/Commands/FBSessionCommands.m
@@ -15,6 +15,7 @@
 #import "FBApplication.h"
 #import "XCUIDevice.h"
 #import "XCUIDevice+FBHealthCheck.h"
+#import "XCUIDevice+FBHelpers.h"
 
 @implementation FBSessionCommands
 
@@ -84,6 +85,7 @@
       @"ios" :
         @{
           @"simulatorVersion" : [[UIDevice currentDevice] systemVersion],
+          @"ip" : [XCUIDevice sharedDevice].fb_wifiIPAddress,
         },
       @"build" :
         @{


marekcirkos:
Isn't this redundant? If you can hit status endpoint you already know it's address right?

imurchie:
For our use case, we use `iproxy` to proxy over USB. And have recently found it much more reliable to just ping WDA instead of reading the logs, which presents a host of problems. But it means we never know the IP, in case we need to switch on the fly from `iproxy` to straight http.

marekcirkos:
@imurchie Can you try using health check endpoint instead? https://github.com/facebook/WebDriverAgent/blob/master/WebDriverAgentLib/Commands/FBSessionCommands.m#L33

imurchie:
Do you mean adding the `ip` to the health check?

marekcirkos:
I assume you want to determine if WDA has booted? That's why you parse logs? Instead you could try to hit `/healthcheck` endpoint after you start WDA or this is problematic as well?

imurchie:
Oh. We were parsing the logs, and getting the IP from there, as well as determining that things are running. Now we hit `/status`. This works perfectly well (if `/healthcheck` is better, I can change it to that and see).

But this way we do not know what the ip of the device is, and there is sometimes a need to fall back to a mode that uses the IP. It seems like device information of the sort that `/status` returns (udid, etc.), so having it returned there was logical, and of little overhead.

marekcirkos:
I see. If you want we can add it, no problem. However why not keep whole connection via USB? This way you don't enforce same WiFi on device or even can test device in AirPlane mode.

imurchie:
We generally do. But there are times when the USB proxy fails, so we want to bypass it.

marekcirkos:
Makes sense. Landing.

PR #456, opened 2017-01-19
Fix parsing of predicate string

imurchie:
This was not able to get the actual suffix, since it modified the property name as it went. Instead it should pull the substrings from the original property name. Also `substringFromIndex` is inclusive, so we need to go from one past the position of the dot.

### DIFF
--- a/WebDriverAgentLib/Categories/NSExpression+FBFormat.m
+++ b/WebDriverAgentLib/Categories/NSExpression+FBFormat.m
@@ -17,14 +17,14 @@
   if ([input expressionType] != NSKeyPathExpressionType) {
     return input;
   }
-  NSString *actualPropName = [input keyPath];
-  NSUInteger dotPos = [actualPropName rangeOfString:@"."].location;
+  NSString *propName = [input keyPath];
+  NSUInteger dotPos = [propName rangeOfString:@"."].location;
   if (NSNotFound != dotPos) {
-    actualPropName = [actualPropName substringToIndex:dotPos];
-    NSString *suffix = [actualPropName substringFromIndex:dotPos];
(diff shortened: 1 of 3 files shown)


WebDriverAgentLib/Categories/NSExpression+FBFormat.m
```diff
@@ -17,14 +17,14 @@ + (instancetype)fb_wdExpressionWithExpression:(NSExpression *)input
...
-  NSString *actualPropName = [input keyPath];
-  NSUInteger dotPos = [actualPropName rangeOfString:@"."].location;
+  NSString *propName = [input keyPath];
+  NSUInteger dotPos = [propName rangeOfString:@"."].location;
   if (NSNotFound != dotPos) {
-    actualPropName = [actualPropName substringToIndex:dotPos];
-    NSString *suffix = [actualPropName substringFromIndex:dotPos];
+    NSString *actualPropName = [propName substringToIndex:dotPos];
+    NSString *suffix = [propName substringFromIndex:(dotPos + 1)];
     return [NSExpression expressionForKeyPath:[NSString stringWithFormat:@"%@.%@", [FBElementUtils wdAttributeNameForAttributeName:actualPropName], suffix]];
```
mykola-mokhnach: could you please add unit test for such case? The tests container is located in UnitTests folder, NSExpressionFBFormatTests.m file.

marekcirkos: +1

imurchie: I've updated the existing unit tests to make useful assertions.

PR #464, opened 2017-01-25
Alternative visibility detection approach

mykola-mokhnach:
I hope this is the last commit, which is supposed to finally fix this annoying issue described in https://github.com/facebook/WebDriverAgent/issues/372

Yes, it has some drawbacks and I have documented them. But it works for our automation and for most of the people who tried it. Also, I've left a workaround for the old visibility detection method by setting the corresponding env variable, if one prefers preciseness over stability. 

@imurchie @jlipps FYI


jlipps:
Change seems reasonable, let's just document the hell out of it on the Appium side as well, and if it's going to be a breaking change for anyone we need to be careful with when we pull this into Appium.

Also this PR needs some serious squashing :-)

PR #466, opened 2017-01-25
Make the visibility property and XML attribute behaviour configurable via /wda/settings endpoint

mykola-mokhnach:
I hope this is the last commit, which is supposed to finally fix this annoying issue described in #372

Now it is possible to switch visibility detection algorithm explicitly by setting the corresponding environment variable. So, in case there are freezing issues, it's always possible to use less accurate, but stable approach. 
Same with XML/XPath. One may just enable adding 'visible' attribute to XML nodes if he really uses it for XPath lookup.

All new features are disabled by default (which simulates the current _master_ behaviour), so the proposed change should not affect any existing tests.

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.h
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.h
@@ -14,7 +14,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface XCUIElement (FBIsVisible)
 
-/*! Whether or not the element is visible */
+/*! Whether or not the element is visible.
+ The value for this property is equal to
+ element snapshot's fb_isVisible property.
+ */
 @property (atomic, readonly) BOOL fb_isVisible;
 
 @end
@@ -22,7 +25,18 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface XCElementSnapshot (FBIsVisible)
 
-/*! Whether or not the element is visible */
+/*! Whether or not the element is visible.
+ The current visibility detection algorithm gives almost 100% accuracy,
+ but in some situations it can cause unexpected delays and
+ failures with "Error copying attributes -25202" record in logs.
+ See https://github.com/facebook/WebDriverAgent/issues/372 for more details.
+ 
+ Set 'useAlternativeVisibilityDetection' setting variable to YES
+ using PUT 'wda/settings' API call if you want to use visibility detection based on snapshot frame
+ analysis instead. This method does not cause unexpected test freezes,
+ although it is not able to properly detect visibility value for some UI elements,
+ which are present in the UI tree, but are not really visible in the app interface.
+ */
 @property (atomic, readonly) BOOL fb_isVisible;
 
 @end
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
@@ -9,6 +9,8 @@
 
 #import "XCUIElement+FBIsVisible.h"
 
+#import "FBConfiguration.h"
+#import "FBMathUtils.h"
 #import "XCElementSnapshot+FBHelpers.h"
 #import "XCTestPrivateSymbols.h"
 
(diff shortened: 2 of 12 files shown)


jlipps:
👍  good to keep current behavior the default

limitcracker:
Hi all,

I went to **XCode > Build** Settings and I set the environmental variable **ALTERNATIVE_VISIBILITY_DETECTION** to **YES** but we got much more SOCKET TIMEOUT errors. It stops unexpectedly while writing usually. Same happens while setting **INCLUDE_VISIBLE_ATTRIBUTE**.

image

`org.openqa.selenium.WebDriverException: An unknown server-side error occurred while processing the command. Original error: Could not proxy command to remote server. Original error: Error: ESOCKETTIMEDOUT (WARNING: The server did not provide any stacktrace information)`

Any other ways I might try?

marekcirkos:
@mykola-mokhnach thank you for your time and awesome support for WDA. Unfortunate I do not like this approach. 
1) Configuring WDA via env and making direct env checks all over the place might not be best choice. I think better approach would be to:
     a) hide that behind `FBConfiguration`. This way if we change way we configure it will not influence rest of the code base.
    b) instead of env maybe pass that on session start, same way we handle 'shouldWaitForQuiescence'?

2) We should add bunch of tests cases for visibility checks, like offscreen elements, etc.

3) Does it make sense to split ALTERNATIVE_VISIBILITY_DETECTION and INCLUDE_VISIBLE_ATTRIBUTE? I would assume that on most of the cases you would like to change both?

Probably that was your intention, but I want to note that we should aim to support only one visibility check. Therefor behaviour switch should be only temporary.

mykola-mokhnach:
> @mykola-mokhnach thank you for your time and awesome support for WDA. Unfortunate I do not like this approach.

Yup, the solution is far from the ideal. I'm just trying to implement in the most flexible way.

> Configuring WDA via env and making direct env checks all over the place might not be best choice. I think better approach would be to:
> a) hide that behind FBConfiguration. This way if we change way we configure it will not influence rest of the code base.
> b) instead of env maybe pass that on session start, same way we handle 'shouldWaitForQuiescence'?

We can do that, but what if it is necessary to switch the behaviour temporarily while a test is running? Let say there are some elements on some views, which cause "cannot read attributes" issue, but not all of them. This means that one can only switch the thing temporarily to workaround the problem and then enable it back for precise visibility detection.
In case this parameter is passed once upon WDA session initialisation there will be no such possibility.

> We should add bunch of tests cases for visibility checks, like offscreen elements, etc.

Is it just to confirm known problems of the alternative visibility detection?
Also, as you might currently notice, I'm more into low-level/logical stuff development. iOS UI developemnt is a completely new area to me. So you can expect a lot of s%$t code there after it's done :-/

> Does it make sense to split ALTERNATIVE_VISIBILITY_DETECTION and INCLUDE_VISIBLE_ATTRIBUTE? I would assume that on most of the cases you would like to change both?

**ALTERNATIVE_VISIBILITY_DETECTION** is responsible for overall elements visibility detection and **INCLUDE_VISIBLE_ATTRIBUTE** is just about including the \@visbile attribute into the XML tree, which WDA generates from application UI. Probably, I had to select a better name for the second one to make it more clear %)

jlipps:
One comment on this: Appium supports a Settings API for params which might change in the course of a single session. Maybe it would be better to implement this API rather than switch based on env vars?

From a client perspective it would look something like this:

    driver.setting('visibilityStrategy', 'alternate');

mykola-mokhnach:
@jlipps there is no other way for the external client to change env vars for iOS process while  it is running. This API that you mentioned will have to trigger the same WDA method I've already implemented.

jlipps:
maybe I'm not understanding, @mykola-mokhnach. Couldn't WDA itself implement this API? Isn't what we're trying to do to get a flag into WDA so that it can change it's behavior based on this flag? And since WDA is a JSONWP server, couldn't we use that mechanism to send our intention to switch the flag? (Which it keeps in some global variable of its own, perhaps).

mykola-mokhnach:
@jlipps 
Yeah, I got your point. Actually, this would look more promising if one just could create a global singleton to keep these settings and then modify them on the fly (in comparison to "raw" env variables access). Can you please provide a link to the corresponding spec?

jlipps:
@mykola-mokhnach we have a pretty comprehensive doc here: https://github.com/appium/appium/blob/master/docs/en/advanced-concepts/settings.md

mykola-mokhnach:
@jlipps @marekcirkos How do you like it now? I suppose this is closer to what we discussed.

senevit4:
Anything new about this PR?

marekcirkos:
@senevit4 I don't want to land this. I working on replacement of current `is_visible` check.

mykola-mokhnach:
@marekcirkos I'd be happy to close this PR
However there is one question - how do you think, would it be useful to be able to change FBConfiguration settings in runtime like it is implemented in this PR or we keep it as is and only allow to set it on session startup?

marekcirkos:
I strongly want to push for one solution. Configuration is there just as a backup in case "new way" is completely broken.
I think that if you allow switch in the middle it will encourage to depend on ability to switch and we will end-up with both implementations.

mykola-mokhnach:
OK, thanks for the reply

PR #470, opened 2017-01-30
Retry alert button tap by coordinates in non-portrait mode if “classic” tap didn’t work

mykola-mokhnach:
It seems like there is difference between how XCTest handles the UI when a system or self-generated alert is visible and the current UI orientation is not portrait. This PR is trying to workaround the bug by doing "classic" tap on alert button and then retrying this by tapping on coordinates if the previous method failed. Otherwise it is not possible to accept system alert properly without manually changing interface orientation to portrait every time such alert appears on the screen.

Addresses https://github.com/appium/appium/issues/7772

### DIFF
--- a/WebDriverAgentLib/FBAlert.m
+++ b/WebDriverAgentLib/FBAlert.m
@@ -123,7 +123,14 @@ NSString *const FBAlertObstructingElementException = @"FBAlertObstructingElement
       withDescriptionFormat:@"Failed to find accept button for alert: %@", alertElement]
      buildError:error];
   }
-  return [defaultButton fb_tapWithError:error];
+  [defaultButton tap];
+  if (defaultButton.exists && defaultButton.application.interfaceOrientation != UIDeviceOrientationPortrait) {
+    // Tap method invokation might not work as expected in non-portrait mode
+    // due to a known XCTest bug. That is why we need to retry it using tap
+    // by coordinates if the alert is still visible
+    return [defaultButton fb_tapWithError:error];
+  }
+  return YES;
 }
 
 - (BOOL)dismissWithError:(NSError **)error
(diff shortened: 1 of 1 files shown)


WebDriverAgentLib/FBAlert.m
```diff
@@ -123,7 +123,14 @@ - (BOOL)acceptWithError:(NSError **)error
       withDescriptionFormat:@"Failed to find accept button for alert: %@", alertElement]
      buildError:error];
   }
-  return [defaultButton fb_tapWithError:error];
+  [defaultButton tap];
```
marekcirkos: 1) You trigger tap twice, which might bring unexpected behaviour. 
2) There is no tests that covers that issue.

mykola-mokhnach: 1. Yes, since I don't see any other possibilities. Also, the code will tap a button for the second time only if it is still visible after the first tap and if the screen orientation is different from portrait, which makes this less likely to happen.
Maybe you have any idea on how to distinguish system security alert from a custom alert? 

2. The test needs to be executed on iPad only and it is necessary to invoke system security alert, which requires changes in our test integration app design. Would it be easy to implement it with the current stuff? 

The easiest solution is just to keep it as is and say people to change to landscape if they need to handle system alerts in landscape. It's anyway XCTest bug }:-@

marekcirkos:
This is hacky solution. I'd rather fix underlying problem there. I would assume coordinates of system alerts are different. Is that correct?

We definitely need test to cover that. We already run tests on iPhones & iPads and have tests on system and non-system alerts, so it should definitely be possible.

mykola-mokhnach:
[Screenshot](#Screenshot)

As you can see here XCTest "corrupts" element element coordinates if such alert is visible, but somehow **XCUIElement** methods (_tap_ for example) can handle that, unlike in the "normal" landscape mode, where we need to recalculate coordinates manually and avoid using  **XCUIElement** methods (we do click by coordinates instead). 

Probably, it would be better to suspend the thing for now. Such workarounds for XCTest bugs are taking too much time and I see it only pays itself back if there is no other way to overcome it.

PR #473, opened 2017-01-31
Add separate API call for on-screen keyboard dismissal

mykola-mokhnach:
We have _hideKeyboard_ API call in Appium, which is quite hacky for iOS right now. And since there is a special method in _XCUIApplication_ class for exactly such purpose, then the stuff should be implemented in the right way.
@imurchie @jlipps FYI

### DIFF
--- a/WebDriverAgentLib/Commands/FBCustomCommands.m
+++ b/WebDriverAgentLib/Commands/FBCustomCommands.m
@@ -24,6 +24,7 @@
 #import "XCUIApplication+FBHelpers.h"
 #import "XCUIDevice+FBHelpers.h"
 #import "XCUIElement.h"
+#import "XCUIElement+FBIsVisible.h"
 #import "XCUIElementQuery.h"
 
 @implementation FBCustomCommands
@@ -76,16 +77,20 @@
 {
   [request.session.application dismissKeyboard];
   NSError *error;
-  BOOL isKbdNotPresent =
+  NSString *errorDescription = @"The keyboard cannot be dismissed. Try to dismiss it in the way supported by your application under test.";
+  if ([UIDevice.currentDevice userInterfaceIdiom] == UIUserInterfaceIdiomPhone) {
+    errorDescription = @"The keyboard on iPhone cannot be dismissed because of a known XCTest issue. Try to dismiss it in the way supported by your application under test.";
+  }
+  BOOL isKeyboardNotPresent =
   [[[[FBRunLoopSpinner new]
      timeout:5]
-    timeoutErrorMessage:@"The keyboard cannot be dismissed. Try to use custom approach instead."]
+    timeoutErrorMessage:errorDescription]
    spinUntilTrue:^BOOL{
      XCUIElement *foundKeyboard = [[FBApplication fb_activeApplication].query descendantsMatchingType:XCUIElementTypeKeyboard].element;
-     return !(foundKeyboard.exists && foundKeyboard.hittable);
+     return !(foundKeyboard.exists && foundKeyboard.fb_isVisible);
    }
    error:&error];
-  if (!isKbdNotPresent) {
+  if (!isKeyboardNotPresent) {
     return FBResponseWithError(error);
   }
   return FBResponseWithOK();
--- a/WebDriverAgentLib/Commands/FBCustomCommands.m
+++ b/WebDriverAgentLib/Commands/FBCustomCommands.m
@@ -18,6 +18,7 @@
 #import "FBResponsePayload.h"
 #import "FBRoute.h"
 #import "FBRouteRequest.h"
+#import "FBRunLoopSpinner.h"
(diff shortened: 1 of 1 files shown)


WebDriverAgentLib/Commands/FBCustomCommands.m
```diff
@@ -70,4 +72,23 @@ + (NSArray *)routes
   return FBResponseWithOK();
 }
 
++ (id<FBResponsePayload>)handleDismissKeyboardCommand:(FBRouteRequest *)request
+{
+  [request.session.application dismissKeyboard];
```
marekcirkos: I am pretty sure this will not work on iPhones.

mykola-mokhnach: Yes, I tested it and it works merely on iPad (or where **Hide keyboard** button is explicitly present). But it's anyway not our issue.

This was also the reason why I added an error message and keyboard presence verification - to let end users know that the call does not work and they must use custom approaches to hide the keyboard instead.

marekcirkos: I am resistant in landing stuff like this, because this will cause tons of issues complaining about it not working on iPhones. If you really need this. Please check for type of device and make the message super clear, that failure on iPhones is expected.

WebDriverAgentLib/Commands/FBCustomCommands.m
```diff
@@ -70,4 +72,23 @@ + (NSArray *)routes
...
+  NSError *error;
+  BOOL isKbdNotPresent =
```
marekcirkos: Please do not use abbreviations. `isKeyboardPresent`

WebDriverAgentLib/Commands/FBCustomCommands.m
```diff
@@ -70,4 +72,23 @@ + (NSArray *)routes
...
+  [[[[FBRunLoopSpinner new]
+     timeout:5]
+    timeoutErrorMessage:@"The keyboard cannot be dismissed. Try to use custom approach instead."]
```
marekcirkos: This message is to custom. I would make check on device and make clear message on iPhones that failure is expected.

WebDriverAgentLib/Commands/FBCustomCommands.m
```diff
@@ -70,4 +72,23 @@ + (NSArray *)routes
...
+   spinUntilTrue:^BOOL{
+     XCUIElement *foundKeyboard = [[FBApplication fb_activeApplication].query descendantsMatchingType:XCUIElementTypeKeyboard].element;
+     return !(foundKeyboard.exists && foundKeyboard.hittable);
```
marekcirkos: Why not check for visibility?

mykola-mokhnach: because this approach is supposed to be faster (_fb_isVisible_ requires XCUIElement instance to be resolved, which is an expensive operation) and will not potentially cause "cannot read attributes" issue, which we all like so much :).

WebDriverAgentLib/Commands/FBCustomCommands.m
mykola-mokhnach: ok will do

WebDriverAgentLib/Commands/FBCustomCommands.m
mykola-mokhnach: sure, thanks for the suggestion. 
And yes, we need this method in Appium, since we had the _hideKeyboard_ function in our API from the beginning and right now it became hacky and slow. I'd like to just have one native call there instead.

WebDriverAgentLib/Commands/FBCustomCommands.m
mykola-mokhnach: did you mean _isKeyboardNotPresent_ ?
Or you want me to change the block itself, so it can also expect False values?

marekcirkos: Both work for me. Personally I find `isKeyboardPresent` easier to read (it is easy to skip not in the middle) , but if you prefer `isKeyboardNotPresent` I am fine with it.

WebDriverAgentLib/Commands/FBCustomCommands.m
marekcirkos: `exists` or/and `hittable` will resolve element anyway. I understand 'cannot read attributes' argument, but this is only because is_visible check is broken, which we need to fix anyway. I should have some time next week to work on it, but I'd rather keep code simpler and unified. 

You could add that if to is_visible check, to filter obvious cases :)

WebDriverAgentLib/Commands/FBCustomCommands.m
mykola-mokhnach: changed to isKeyboardNotPresent

PR #511, opened 2017-03-10
Add nil check to FBFirstNonEmptyValue() macro.

Ysoko:
Fixes issue introduced by 7113842e09f3ea136904fadd56011fe0c28d8f3a, where element.wdValue is nil and instead of falling back on element.wdLabel, nil is assigned to text instead.


marekcirkos:
Wow, first time I see nil returned by XCTest. Thanks!

Issue #481, opened 2017-02-13
When setting value, connection is sometimes dropped


imurchie:
WebDriverAgent sometimes drops the connection with the client. While this can happen randomly with any command, it seems to happen most on resource-constrained systems (when there is low memory, etc.) while sending a `POST` request to `/element/:uuid/value`. This is either felt as the socket hanging up (if waiting for a response) or an ECONNREFUSED error (if sending a request.

A sample of the sort of thing that is begin done (in this particular case, this is about 4 minutes into a session):
```
[debug] [JSONWP Proxy] Proxying [GET /wd/hub/session/465bb6be-c916-491d-8e4c-5603f38961df/element/57588822-2EE7-4CEA-A508-073569758610/displayed] to [GET http://localhost:8100/session/3A134330-F9E1-47A8-8B58-5607AD399477/element/57588822-2EE7-4CEA-A508-073569758610/displayed] with body: {}
[debug] [JSONWP Proxy] Got response with status 200: "{\n  \"value\" : true,\n  \"sessionId\" : \"3A134330-F9E1-47A8-8B58-5607AD399477\",\n  \"status\" : 0\n}"

[debug] [JSONWP Proxy] Proxying [GET /element/57588822-2EE7-4CEA-A508-073569758610/attribute/type] to [GET http://localhost:8100/session/3A134330-F9E1-47A8-8B58-5607AD399477/element/57588822-2EE7-4CEA-A508-073569758610/attribute/type] with no body
[debug] [JSONWP Proxy] Got response with status 200: "{\n  \"value\" : \"XCUIElementTypeTextField\",\n  \"sessionId\" : \"3A134330-F9E1-47A8-8B58-5607AD399477\",\n  \"status\" : 0\n}"

[debug] [JSONWP Proxy] Proxying [POST /element] to [POST http://localhost:8100/session/3A134330-F9E1-47A8-8B58-5607AD399477/element] with body: {"using":"class name","value":"XCUIElementTypeKeyboard"}
2017-02-10 22:50:01:914 - [debug] [JSONWP Proxy] Got response with status 200: {"value":{"ELEMENT":"FB16F40C-F5A2-4C8F-B813-5E3CC54F12E1","type":"XCUIElementTypeKeyboard","label":null},"sessionId":"3A134330-F9E1-47A8-8B58-5607AD399477","status":0}

[...]

PR #513, opened 2017-03-14
Add way to use XCTest typing method

imurchie:


imurchie:
Whoops. This is a work in progress. Ignore.

marekcirkos:
@imurchie Just curious, what are you working on?

jlipps:
We've noticed that WDA goes away or crashes on some systems (perhaps low-resource systems) when using send keys. @imurchie is working on a different way to set text that doesn't have this adverse side effect.

imurchie:
Yes. This is work in progress to try to alleviate https://github.com/facebook/WebDriverAgent/issues/481 which I have tracked down to the [typing](https://github.com/facebook/WebDriverAgent/blob/master/WebDriverAgentLib/Utilities/FBKeyboard.m#L33). I'm essentially trying out `XCUIElement` method `typeText`. Seems to work except in some situations things are typed too fast and the gestures run into each other and fail.

PR #522, opened 2017-03-24
Extend class chain search to use predicates

mykola-mokhnach:
I think that's all we can do to avoid xpath usage and still keep the lookup fast.


marekcirkos:
I can't land it now, not sure if there is conflict or what. I'll get back to it next week, when I am back :)

mykola-mokhnach:
Probably, that's even better, that we didn't land it, since I decided to change the default enclosing marker for predicate expression. This will allow to avoid extra quotes escaping inside class chain queries ;)

PR #532, opened 2017-03-31
Silence deprecation error for openUrl

imurchie:
A quick down-and-dirty fix for the inability to build in Xcode 8.3 because of the following error:
```
/Users/isaac/code/isaac-WDA/WebDriverAgentLib/Commands/FBSessionCommands.m:56:43: error: 'openURL:' is deprecated: first
      deprecated in iOS 10.0 - Please use openURL:options:completionHandler: instead [-Werror,-Wdeprecated-declarations]
  if (![[UIApplication sharedApplication] openURL:url]) {
                                          ^
In module 'XCTest' imported from /Users/isaac/code/isaac-WDA/PrivateHeaders/XCTest/XCUIApplication.h:7:
In module 'UIKit' imported from /Applications/Xcode-8.3.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/Library/Frameworks/XCTest.framework/Headers/XCUIElement.h:8:
/Applications/Xcode-8.3.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator10.3.sdk/System/Library/Frameworks/UIKit.framework/Headers/UIApplication.h:125:1: note:
      'openURL:' has been explicitly marked deprecated here
- (BOOL)openURL:(NSURL*)url NS_DEPRECATED_IOS(2_0, 10_0, "Please use openURL:options:completionHandler: instead"...
^
1 error generated.

** TEST BUILD FAILED **
```

### DIFF
--- a/WebDriverAgentLib/Commands/FBSessionCommands.m
+++ b/WebDriverAgentLib/Commands/FBSessionCommands.m
@@ -53,9 +53,12 @@
       [NSString stringWithFormat:@"%@ is not a valid URL", url]
     );
   }
+  #pragma clang diagnostic push
+  #pragma clang diagnostic ignored "-Wdeprecated-declarations"
   if (![[UIApplication sharedApplication] openURL:url]) {
     return FBResponseWithErrorFormat(@"Failed to open %@", url);
   }
+  #pragma clang diagnostic pop
   return FBResponseWithOK();
 }
 


WebDriverAgentLib/Commands/FBSessionCommands.m
```diff
@@ -53,6 +53,7 @@ + (NSArray *)routes
       [NSString stringWithFormat:@"%@ is not a valid URL", url]
     );
   }
+  #pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
mykola-mokhnach: wouldn't it be more correct to replace the deprecated declaration with the up-to-date one?(openURL:options:completionHandler:)

imurchie: Yeah, except it doesn't work.

marekcirkos: 1. You should use `#pragma clang diagnostic push` and `#pragma clang diagnostic pop` to make it local.
2. I agree with @mykola-mokhnach. You should add `if` that checks for the new API and use old API in case it is not available. Compiler SHOULD be smart enough to not generate warning

mykola-mokhnach:
@imurchie There are predefined macroses for system version comparison in FBMacros.h

imurchie:
Didn't know about the push/pop.

The compiler still needed the pragma marks when I tried to do the split, based on whether the API call was there. But in the end it didn't matter, since I couldn't get the new call to work at all.

I don't have time to try to get it to work, also, since Appium does not actually use the method. It just makes it impossible to build.

imurchie:
The API level checks would be less reliable than checking if the system responds to `openURL:options:completionHandler:`.

imurchie:
I've updated to make the silencing local.

PR #534, opened 2017-04-02
Replace invalid xml characters to avoid unexpected crashes

mykola-mokhnach:
This resolves the issue when application under test contains some characters in resources, which are unsupported in XML, and causes WDA to throw an exception, for example https://github.com/appium/appium/issues/8084

With this patch WDA will be able to build page source tree properly by simply cutting off all unsupported characters.

### DIFF
--- a/WebDriverAgentLib/Categories/NSString+FBXMLSafeString.h
+++ b/WebDriverAgentLib/Categories/NSString+FBXMLSafeString.h
@@ -1,0 +1,28 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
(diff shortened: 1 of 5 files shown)


WebDriverAgentLib/Categories/NSString+FBXMLSafeString.h
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ Method used to normalize a string before passing it to XML document
+ 
+ @param replacement The string to be used as a replacement for invalid XML characters
+ @return The string where all characters, which are not members of
+         XML Character Range definition (http://www.w3.org/TR/2008/REC-xml-20081126/#charsets),
+         are replaced
+ */
+- (NSString *)xmlSafeStringWithReplacement:(NSString *)replacement;
```
marekcirkos: This is 'public' class you should prefix it eg. `fb_ xmlSafeStringWithReplacement`

marekcirkos:
Otherwise looks good to me

PR #539, opened 2017-04-05
Switch from class property to methods

imurchie:
Some recent changes use class properties, which break building with Xcode 7. This exposes the same API without the new syntax.

### DIFF
--- a/WebDriverAgentLib/Utilities/FBConfiguration.h
+++ b/WebDriverAgentLib/Utilities/FBConfiguration.h
@@ -17,10 +17,12 @@ NS_ASSUME_NONNULL_BEGIN
 @interface FBConfiguration : NSObject
 
 /*! If set to YES will ask TestManagerDaemon for element visibility */
-@property (class, nonatomic, assign) BOOL shouldUseTestManagerForVisibilityDetection;
++ (void)setShouldUseTestManagerForVisibilityDetection:(BOOL)value;
++ (BOOL)shouldUseTestManagerForVisibilityDetection;
 
 /* The maximum typing frequency for all typing activities */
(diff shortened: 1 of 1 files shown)


marekcirkos:
Do we care about supporting Xcode 7?

imurchie:
Depends on who "we" is. _We_ do!

That being said, if it is a conscious decision to remove support for it, it should be explicit I think.

marekcirkos:
By we I mean Appium and FB. I understand need to support old iOS version, but I don't see reason to support old Xcode.

imurchie:
In general, I agree. Xcode 7.3 has advantages for iOS 9.3 simulator support (performance being one major one). In practical terms, we have users who need to be able to use it until their infrastructure can be upgraded. Alas.

marekcirkos:
There is also one more problem here. XCUITest (and WDA)  is mainly dependent on Xcode version it is build on. Only few improvements depend on iOS version. That means if XCTest framework changes significantly between Xcode versions (which often happen even between small version bumps) it is hard to keep them compiling and working on all of them. Mainly because private API changes.

marekcirkos:
What I am saying is. If you really care about this change I can marge it. However there is big chance that other stuff is broken or will break soon.

imurchie:
Yes, things break relatively often. Some things we rewrite before compiling so they work (like private header changes).  I guess we could do that here, too.

marekcirkos:
What do you mean?

marekcirkos:
@imurchie? Still want to land this?

imurchie:
Still do, yes. Right now we're working off a fork because of this (and one other thing we have).

PR #546, opened 2017-04-14
configure text setting in project

fabb:
so the 2-spaces default gets applied independently of global xcode settings. helps contributors to keep text formatted properly for this project.


marekcirkos:
Thank you!

PR #549, opened 2017-04-19
Make the offset configurable

mykola-mokhnach:
This PR adds an optional _offset_ attribute to _/wda/pickerwheel/:uuid/select_ API call. This is done, because the default offset, which was previously equal to 0.2 may not work as expected for some picker wheel implementations and make the wheel to switch more than one value at once. Now this behaviour is configurable.

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBPickerWheel.h
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBPickerWheel.h
@@ -16,20 +16,30 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  Selects the next available option in Picker Wheel
  
+ @param offset the value in range [0.01, 0.5]. It defines how far from picker
+   wheel's center the click should happen. The actual distance is culculated by
+   multiplying this value to the actual picker wheel height. Too small offset value
+   may not change the picker wheel value and too high value may cause the wheel to switch
+   two or more values at once. Usually the optimal value is located in range [0.15, 0.3]
  @param error returns error object if there was an error while selecting the
    next picker value
  @return YES if the current option has been successfully switched. Otherwise NO
  */
-- (BOOL)fb_selectNextOptionWithError:(NSError **)error;
+- (BOOL)fb_selectNextOptionWithOffset:(CGFloat)offset error:(NSError **)error;
 
 /**
  Selects the previous available option in Picker Wheel
  
- @param error returns error object if there was an error while selecting the 
+ @param offset the value in range [0.01, 0.5]. It defines how far from picker
+   wheel's center the click should happen. The actual distance is culculated by
+   multiplying this value to the actual picker wheel height. Too small offset value
+   may not change the picker wheel value and too high value may cause the wheel to switch
+   two or more values at once. Usually the optimal value is located in range [0.15, 0.3]
+ @param error returns error object if there was an error while selecting the
    previous picker value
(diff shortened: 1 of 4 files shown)


WebDriverAgentLib/Categories/XCUIElement+FBPickerWheel.h
```diff
@@ -16,20 +16,30 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @param offset the value in range [0.01, 0.5]. It defines how far from picker
+   wheel's center the click should happen. The actual distance is culculated by
+   multiplying this value to the actual picker wheel height. Too small offset value
+   may not change the picker wheel value and too high value may cause the wheel to switch
+   two or more values at once. Usually the optimal value is located in range [0.15, 0.3]
+ @param error returns error object if there was an error while selecting the
    previous picker value
  @return YES if the current option has been successfully switched. Otherwise NO
  */
-- (BOOL)fb_selectPreviousOptionWithError:(NSError **)error;
```
marekcirkos: At this point probably we should start document custom endpoints on Wiki :)

WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -379,12 +379,19 @@ + (NSArray *)routes
     return FBResponseWithErrorFormat(@"The element is expected to be a valid Picker Wheel control. '%@' was given instead", element.wdType);
   }
   NSString* order = [request.arguments[@"order"] lowercaseString];
+  CGFloat offset = (CGFloat)0.2;
```
marekcirkos: I don't think you need to cast it. It will do it for you.

WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -379,12 +379,19 @@ + (NSArray *)routes
...
+  if (request.arguments[@"offset"]) {
+    offset = (CGFloat)[request.arguments[@"offset"] doubleValue];
```
marekcirkos: Same here. Type is known to be CGFloat so compiler will do that for you.

WebDriverAgentLib/Categories/XCUIElement+FBPickerWheel.h
mykola-mokhnach: will add this to Wiki as soon as the PR is merged

WebDriverAgentLib/Commands/FBElementCommands.m
mykola-mokhnach: last time the compiler didn't like that w/o explicit typecast, since CGFloat is a typedef for float on x86 and for double on x64

WebDriverAgentLib/Commands/FBElementCommands.m
mykola-mokhnach: OK, let's see if it still keeps the compiler happy

WebDriverAgentLib/Categories/XCUIElement+FBPickerWheel.h
marekcirkos: We need to create one that is proper documentation. Current one is just use sample.

mykola-mokhnach: I would be happy to add all the necessary details if you can provide a template for it

butterflyingdog:
ok

PR #557, opened 2017-04-27
Rename ‘shown’ attribute to ‘visible’ to keep the backward compatibility

mykola-mokhnach:
subj

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
@@ -11,7 +11,6 @@
 
 #import "FBApplication.h"
 #import "FBConfiguration.h"
-#import "FBLogger.h"
 #import "FBMathUtils.h"
 #import "XCElementSnapshot+FBHelpers.h"
 #import "XCTestPrivateSymbols.h"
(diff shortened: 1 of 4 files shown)


WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
```diff
@@ -43,8 +43,16 @@ - (BOOL)fb_isVisible
...
   BOOL rectIntersects = CGRectIntersectsRect(self.visibleFrame, screenFrame);
-  BOOL isActionable = CGRectContainsPoint(app.frame, self.hitPoint);
-  return rectIntersects && isActionable;
+  @try {
+    BOOL isActionable = CGRectContainsPoint(app.frame, self.hitPoint);
+    return rectIntersects && isActionable;
+  } @catch (NSException *e) {
+    if ([e.reason containsString:@"attempt to insert nil object"]) {
+      // This is to workaround known XCTest issue https://github.com/facebook/WebDriverAgent/issues/542
+      return rectIntersects;
```
marekcirkos: Not sure if this is a good idea. That will be inconsistent.
Can we:
1) Log that it happen with some additional info for diagnostics
2) Return hard NO?

mykola-mokhnach: OK, point taken

WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
```diff
@@ -43,8 +44,17 @@ - (BOOL)fb_isVisible
   CGSize screenSize = FBAdjustDimensionsForApplication(app.frame.size, (UIInterfaceOrientation)[XCUIDevice sharedDevice].orientation);
   CGRect screenFrame = CGRectMake(0, 0, screenSize.width, screenSize.height);
   BOOL rectIntersects = CGRectIntersectsRect(self.visibleFrame, screenFrame);
-  BOOL isActionable = CGRectContainsPoint(app.frame, self.hitPoint);
-  return rectIntersects && isActionable;
+  @try {
+    BOOL isActionable = CGRectContainsPoint(app.frame, self.hitPoint);
+    return rectIntersects && isActionable;
+  } @catch (NSException *e) {
```
marekcirkos: I am sorry for the problem, but 
1) Can you split this diff to `shown` -> `visible` change and another one with `try`.
2) On 'try' diff add check whether you want to handle error on that manner. Same as we have `shouldUseTestManagerForVisibilityDetection` (default it to whatever you think is best). It would be easier for me to work on the fix for it.

mykola-mokhnach: ok will do

PR #559, opened 2017-04-28
Add visibility detection workaround for the known XCTest exception

mykola-mokhnach:
It is better to return NO as visibility value rather than to crash the whole WDA for some special cases.
See https://github.com/facebook/WebDriverAgent/issues/542 for more details


marekcirkos:
I got few diffs coming that do similar, but adds few tests and wraps this so that we can re-use it later.

PR #566, opened 2017-05-03
Change name parameter to be an argument

mykola-mokhnach:
I think this is against a spec to have the **name** defined as a parameter in the API calls instead of an argument.


mykola-mokhnach:
@imurchie FYI

marekcirkos:
This is out of spec anyway :)

Issue #553, opened 2017-04-26
JSONWP command "runAppInBackground" returns error "Failed to find scrollable visible parent with 2 visible children"


antonys1:
Via Appium the JSONWP command "runAppInBackground" results in an error.

**WebDriverAgent error received via Appium logs:**
> [debug] [JSONWP Proxy] Got response with status 200: {"value":"Error Domain=com.facebook.WebDriverAgent Code=1 "Failed to find scrollable visible parent with 2 visible children" UserInfo={NSLocalizedDescription=Failed to find scrollable visible parent with 2 visible children}","sessionId":"98821523-723A-4DE9-AFCD-80B9D9896BF4","status":13}

**Environment:**
macOS 10.12.4
Xcode 8.3.2
iOS 10.3 Simulator
Appium Server v1.6.4
Appium Java Client v5.0.0-BETA7

marekcirkos:
Do you care about providing a bit more information? I am not familiar with Appium. What `runAppInBackground` is suppose to do? Can you add system logs? screenshot? anything?

antonys1:
> Do you care about providing a bit more information?

Sure, for starters see [original Appium issue](https://github.com/appium/appium/issues/8323)

> I am not familiar with Appium.

Appium is an open source test automation framework for use with native, hybrid and mobile web apps. It drives iOS, Android, and Windows apps using the WebDriver protocol.

> What runAppInBackground is suppose to do?

It's supposed to send the active app to the background, a duration parameter specifies for how long until the app should eventually return to the foreground.

> Can you add system logs? screenshot? anything?

I already added the pertinent logs from Appium as pointed out by one of the Appium contributors, who decided that this was a WDA bug.

I don't think a screenshot will convey anything useful, it would just show the iOS Simulator running the active app.

I think that covers _anything_ unless you can be more specific?

marekcirkos:
@antonys1 It is definitely WDA issue, I need to find a way to reproduce it. Does it happen sometimes or always?

antonys1:
Always

PR #569, opened 2017-05-05
Improve Springboard icon selection algorithm

mykola-mokhnach:
Also fixes https://github.com/facebook/WebDriverAgent/issues/553

### DIFF
--- a/WebDriverAgentLib/FBSpringboardApplication.m
+++ b/WebDriverAgentLib/FBSpringboardApplication.m
@@ -61,7 +61,7 @@
            timeout:1]
           timeoutErrorMessage:errorDescription]
          spinUntilTrue:^BOOL{
-           return floor(appElement.frame.origin.x) != floor(startXOffset);
+           return fabs(appElement.frame.origin.x - startXOffset) > FLT_EPSILON;
          }
          error:error];
       if (!isSwipeSuccessful) {
--- a/WebDriverAgentLib/FBSpringboardApplication.m
+++ b/WebDriverAgentLib/FBSpringboardApplication.m
@@ -9,12 +9,13 @@
 
 #import "FBSpringboardApplication.h"
 
+#import "FBErrorBuilder.h"
 #import "FBRunLoopSpinner.h"
-#import "FBMacros.h"
 #import "XCElementSnapshot+FBHelpers.h"
 #import "XCElementSnapshot.h"
 #import "XCUIApplication+FBHelpers.h"
 #import "XCUIElement+FBIsVisible.h"
+#import "XCUIElement+FBUtilities.h"
 #import "XCUIElement+FBTap.h"
 #import "XCUIElement+FBScrolling.h"
 #import "XCUIElement.h"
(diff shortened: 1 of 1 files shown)


marekcirkos:
Ok I finally fixed integration tests. Can you please rebase?
Also can you tell what was the primary issue here?

mykola-mokhnach:
@marekcirkos I don't know what is the main cause to be honest. I just wanted to fix the crash for our automation and make the flow a bit simpler and faster. I assume it might be some kind of race condition, since I was not able to reproduce the problem on Simulator,. but only on a real device.

WebDriverAgentLib/FBSpringboardApplication.m
```diff
@@ -36,11 +38,37 @@ + (instancetype)fb_springboard
...
+      withDescriptionFormat:@"Cannot locate Springboard icon for '%@' application", identifier]
+     buildError:error];
+  }
+  // Select the most recent installed application if there are multiple matches
+  XCUIElement *appElement = [matchedAppElements lastObject];
+  if (!appElement.fb_isVisible) {
+    CGRect startFrame = appElement.frame;
+    BOOL shouldSwipeToTheRight = startFrame.origin.x < 0;
+    NSString *errorDescription = [NSString stringWithFormat:@"Cannot scroll to Springboard icon for '%@' application", identifier];
+    do {
```
mykola-mokhnach: because there is a possibility (a very little one, but still) that swipe didn't work and thus the icon will never be visible. this can lead to an endless loop and freeze %)

marekcirkos: Is there anything we can do to fix swipe/scroll then? I really want to avoid code duplication.

mykola-mokhnach: I use the method call, which is provided directly by XCTest. This verification is just a safety net to avoid unexpected freeze. Where do you see code duplication here?

marekcirkos: Whole `do` loop does what swipe till visible does.

mykola-mokhnach: I do remember one more important case - it might be, that the icon of the application, which we want to detect on springboard pane, is an element of a folder and is not represented by a separate dashboard icon. This means it will be always invisible and it does not matter how many swipes are done. The current algorithm will properly detect such situation and will fail, however the algorithm based on "pure" visibility detection will end up in an endless loop

marekcirkos: That is a good point. Tho I am pretty sure you can detect that and handle it probably in separate diff.

mykola-mokhnach: Can you please be more specific about what should be done to have the PR merged?

marekcirkos: I want to merge it as is. We can do further improvements in separate diff

PR #573, opened 2017-05-09
WebdriverAgent change port with parameter

cyildirim:
I edit WebDriverAgentRunner in order to change port which WebDriverAgent runs. I added environment variable.

Main purpose of change port, in Appium side, we trying to run parallel tests on a macOS. With the these both change (https://github.com/appium/appium-xcuitest-driver/pull/435/) it is going to be possible

https://github.com/appium/appium-xcuitest-driver/pull/435/

### DIFF
--- a/WebDriverAgent.xcodeproj/xcshareddata/xcschemes/WebDriverAgentRunner.xcscheme
+++ b/WebDriverAgent.xcodeproj/xcshareddata/xcschemes/WebDriverAgentRunner.xcscheme
@@ -61,6 +61,13 @@
             ReferencedContainer = "container:WebDriverAgent.xcodeproj">
          </BuildableReference>
       </MacroExpansion>
+      <EnvironmentVariables>
+         <EnvironmentVariable
+            key = "USE_PORT"
+            value = "$(WDA_PORT)"
+            isEnabled = "YES">
+         </EnvironmentVariable>
+      </EnvironmentVariables>
       <AdditionalOptions>
       </AdditionalOptions>
    </LaunchAction>


qmfrederik:
Out of curiosity, why do you want to do this? You can only have one instance of the WebDriverAgent running on your device, so there should never be a port conflict.

On the macOS side, you can use iProxy to map device port 8100 to any macOS port

cyildirim:
I try parallel execution with multiple device. So i need run more than one WebDriverAgent instance. I tried it with single WebDriver instance but it failed. After that amendment and two different port (e.g 8100 and 8200) it worked well.

qmfrederik:
If I understand it correctly, you have multiple devices (say two). On each device, you run one instance of the WebDriverAgent. Both devices are connected to your PC.

So you should able to have this setup (that's how we do it):
- WebDriverAgent on device A at port 8100
- WebDriverAgent on device B at port 8100

And then on your mac:
- iproxy 8100 8100 [device A]
- iproxy 8100 8200 [device B]

So if you open http://localhost:8100, you connect to the WebDriverAgent running at port 8100 on device A.
If you open http://localhost:8200, you connect to the WebDriverAgent running at port 8100 on device B.

Does that not work for you?

sergd2005:
I did same thing it works , I dont feel that there is need in port configuration. Besides if u run multiple devices via usb -it is much easier to user proxy and connect to localhost rather then figuring out their ip address in network

marekcirkos:
@cyildirim Are you guys happy with iproxy approach?

cyildirim:
Hi Marek, 

I didn't find opportunity to try it over iproxy for long runs. Today i am going to try it again. If it works i am going to inform you.

cyildirim:
It looks like everything is ok with the iproxy. Thanks for your time guys.

PR #581, opened 2017-05-13
Add animation timeout before getting element snapshot to avoid most occurrences of “Error copying attributes -25202” issues

mykola-mokhnach:
This patch helps to solve "Error copying attributes -25202" error and hardcoded 15-second delay caused by this error by taking element snapshot for XML tree building and XCElement visibility detection while UI animations (usually these are transitions between views or loading screens) are in progress.

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
@@ -22,7 +22,7 @@
 
 - (BOOL)fb_isVisible
 {
-  [self waitUntilSnapshotIsStable];
+  [self fb_waitUntilSnapshotIsStable];
   return self.fb_lastSnapshot.fb_isVisible;
 }
 
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
@@ -13,6 +13,7 @@
 #import "FBConfiguration.h"
 #import "FBMathUtils.h"
 #import "XCElementSnapshot+FBHelpers.h"
+#import "XCUIElement+FBUtilities.h"
 #import "XCTestPrivateSymbols.h"
 #import <XCTest/XCUIDevice.h>
 #import "XCElementSnapshot+FBHitPoint.h"
@@ -21,10 +22,8 @@
 
 - (BOOL)fb_isVisible
 {
-  if (!self.lastSnapshot) {
-    [self resolve];
-  }
-  return self.lastSnapshot.fb_isVisible;
+  [self waitUntilSnapshotIsStable];
+  return self.fb_lastSnapshot.fb_isVisible;
 }
 
 @end
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
@@ -22,7 +22,6 @@
 
 - (BOOL)fb_isVisible
 {
-  [self fb_waitUntilSnapshotIsStable];
   return self.fb_lastSnapshot.fb_isVisible;
 }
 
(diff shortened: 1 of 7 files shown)


mykola-mokhnach:
> waitUntilSnapshotIsStable is better version for resolve
No, it has nothing to do to **resolve**. The trick is to call **waitUntilSnapshotIsStable** before taking XCElement's snapshot to avoid visibility detection errors.

marekcirkos:
Doesn't that slowdown tests on animated apps? This kind of reinvents `_waitForQuiescence` that we override if you don't pass `shouldWaitForQuiescence` while session start.

mykola-mokhnach:
@marekcirkos it should not, since the timer is only invoked once when one tries to get the visibility value of an XCElement instance, get page source or execute xpath search. This is for now the full list of actions where the timer is called since all these methods read visibility value from the taken snapshot. 

This is especially painful for getting source and xpath search, since reading visibility for each node in the snapshot means huge delays, because XCTest waits for 15 seconds for each attribute error it faces.

marekcirkos:
I am kind of worried that this happens on element level. So on animated view we will be waiting on every single element.

marekcirkos:
Nether to say, I don't have better idea yet :/

mykola-mokhnach:
> I am kind of worried that this happens on element level. So on animated view we will be waiting on every single element.

This is not going to happen, before the wait happens BEFORE we get the snapshot and visibility verification itself happens in the snapshot itself. In the worst case one may get an additional single 5-seconds delay while isDisplayed/findElementByXPath/source is called.

mykola-mokhnach:
One more comment regarding animations: they are used mostly for transitions, which means they are short and should not last longer than 1-2 seconds. Having active long-running animations in the UI is not a good UX practice in general.

mykola-mokhnach:
@marekcirkos I verified this approach with our automation framework (more than 300 tests) and it's much more stable now. I don't see attribute errors in logs anymore.

WebDriverAgentLib/Categories/XCUIElement+FBUtilities.m
```diff
@@ -10,12 +10,16 @@
...
 #import "FBAlert.h"
+#import "FBLogger.h"
 #import "FBMathUtils.h"
 #import "FBRunLoopSpinner.h"
+#import "XCAXClient_iOS.h"
 #import "XCUIElement+FBWebDriverAttributes.h"
 
 @implementation XCUIElement (FBUtilities)
 
+const NSTimeInterval ANIMATION_TIMEOUT = 5.0;
```
marekcirkos: static const NSTimeInterval FBANIMATION_TIMEOUT = 5.0;

marekcirkos:
You definitely got app with infinite animations, regardless whether this is a good practice or not. That case would rendered WDA a bit useless, especially when fetching source tree as we would wait 5 sec on every single element.

Maybe we could wait once per request on app element?


Btw. I fixed tests, please rebase :)

mykola-mokhnach:
>  That case would rendered WDA a bit useless, especially when fetching source tree as we would wait 5 sec on every single element.

We do wait once - exactly before taking the whole snapshot.  As you can see the wait is only present in XCElement's visibility verification, but not in XCElementSnapshot's.

WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
```diff
@@ -21,10 +22,7 @@ @implementation XCUIElement (FBIsVisible)
 
 - (BOOL)fb_isVisible
 {
-  if (!self.lastSnapshot) {
```
marekcirkos: Why you want to remove this?

mykola-mokhnach: because _fb_lastSnapshot_ already does it

marekcirkos:
Look good. I need to deploy some stuff, before I can land this.

lavi-xu:
Hi @mykola-mokhnach, @marekcirkos , All,

We have a animated app here, and when we try to fetch any element in same windows by xpath multiple times , it always gives 15s delay for this call.

I don't have we can call developer to change their app, is there a option like shouldWaitForQuiescence  to stop this kind of checking for animate APP?

Thanks for your great helps.

BR.

adudurant:
@marekcirkos  Could you tell me the way to solve the problem？This puzzlement had been haunting my mind for a long time

PR #590, opened 2017-05-18
Add element class to the /source route

qmfrederik:
Adds support for retrieving the underlying element class (such as `UIView` and its derivates) and adding it to the `source`.

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBClass.h
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBClass.h
@@ -1,0 +1,30 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <WebDriverAgentLib/XCElementSnapshot.h>
+#import <WebDriverAgentLib/XCUIElement.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface XCUIElement (FBClass)
+
+/*! Element's class (UIView or derived) */
+@property (nonatomic, readonly, copy) NSString *fb_class;
+
+@end
+
+
+@interface XCElementSnapshot (FBClass)
+
+/*! Element's class (UIView or derived) */
+@property (nonatomic, readonly, copy) NSString *fb_class;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/WebDriverAgentLib/Categories/XCUIElement+FBClass.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBClass.m
@@ -1,0 +1,37 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import "XCUIElement+FBClass.h"
+
+@implementation XCUIElement (FBClass)
+
+- (NSString*)fb_class
+{
+  if (!self.lastSnapshot) {
+    [self resolve];
+  }
+  return [self.lastSnapshot fb_class];
+}
+
+@end
+
+@implementation XCElementSnapshot (FBClass)
+
+- (NSString*)fb_class
+{
+  NSNumber* classId = [NSNumber numberWithInt:5004];
+  NSObject* class = [self.additionalAttributes objectForKey:classId];
+  if([class isKindOfClass:[NSString class]]){
+    return (NSString*)class;
+  } else {
+    return nil;
+  }
+}
+
+@end
--- a/WebDriverAgentLib/Categories/XCUIElement+FBClass.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBClass.m
@@ -30,7 +30,7 @@
   if([class isKindOfClass:[NSString class]]){
     return (NSString*)class;
   } else {
-    return nil;
+    return @"";
   }
 }
 
--- a/WebDriverAgentLib/Categories/XCUIElement+FBWebDriverAttributes.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBWebDriverAttributes.m
@@ -15,6 +15,7 @@
 #import "FBMacros.h"
 #import "XCUIElement+FBAccessibility.h"
 #import "XCUIElement+FBIsVisible.h"
+#import "XCUIElement+FBClass.h"
 #import "XCUIElement.h"
 #import "FBElementUtils.h"
 
@@ -88,6 +89,11 @@
 - (NSString *)wdType
 {
   return [FBElementTypeTransformer stringWithElementType:self.elementType];
+}
+
+- (NSString *)wdClass
+{
+  return FBTransferEmptyStringToNil(self.fb_class);
 }
 
 - (CGRect)wdFrame
--- a/WebDriverAgentLib/Utilities/FBXPath.m
+++ b/WebDriverAgentLib/Utilities/FBXPath.m
@@ -224,14 +224,10 @@ NSString *const XCElementSnapshotXPathQueryEvaluationException = @"XCElementSnap
       return rc;
     }
   }
-  
-  NSNumber* classId = [NSNumber numberWithInt:5004];
-  NSString* className = [element.additionalAttributes objectForKey:classId];
-  
-  if (className) {
-    rc = xmlTextWriterWriteAttribute(writer, BAD_CAST "class", [self.class safeXmlStringWithString:className]);
+  if (element.wdClass) {
+    rc = xmlTextWriterWriteAttribute(writer, BAD_CAST "class", [self.class safeXmlStringWithString:element.wdClass]);
     if (rc < 0) {
-      [FBLogger logFmt:@"Failed to invoke libxml2>xmlTextWriterWriteAttribute(class='%@'). Error code: %d", className, rc];
+      [FBLogger logFmt:@"Failed to invoke libxml2>xmlTextWriterWriteAttribute(class='%@'). Error code: %d", element.wdClass, rc];
       return rc;
     }
   }
--- a/WebDriverAgentLib/Utilities/FBXPath.m
+++ b/WebDriverAgentLib/Utilities/FBXPath.m
@@ -224,6 +224,17 @@ NSString *const XCElementSnapshotXPathQueryEvaluationException = @"XCElementSnap
       return rc;
     }
   }
+  
+  NSNumber* classId = [NSNumber numberWithInt:5004];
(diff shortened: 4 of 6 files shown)


qmfrederik:
@marekcirkos I'm mainly creating the PR to get your feedback at this time. I think getting the class name can be of interest. It's hidden in the additionalAttributes dictionary and the value of 5004 seems to be a constant.

If you want to land this, let me know, and I'd be happy to work with you to get this PR ready to merge.

mykola-mokhnach:
I'd rather add this property info FBElement itself. Also, it would be interesting to know if this magic hack works for all snapshot or it can potentially throw an exception. Is there any documentation about it?

marekcirkos:
Don't we have this info already? All name in inspector come from classes?

qmfrederik:
@marekcirkos Let me double check, until now I only found XCUI Test-related classes, not UIKit classes (such as `UIView` and friends).

qmfrederik:
@marekcirkos Take for example the Facebook login page. For the text field with the user e-mail address, you now get:

```
<XCUIElementTypeTextField type="XCUIElementTypeTextField" value="Email address or phone number" name="username-field" label="Username Field" enabled="true" visible="true" x="28" y="302" width="324" height="44"/>
```

This PR would change that to:

```
<XCUIElementTypeTextField type="XCUIElementTypeTextField" value="Email address or phone number" name="username-field" class="UITextField" label="Username Field" enabled="true" visible="true" x="28" y="302" width="324" height="44"/>
```

so adding `class="UITextField"`. I checked the inspector but I don't think it is based on these classes - rather, it appears to use the XCUIElementTypes?

qmfrederik:
@mykola-mokhnach No, I don't think there's a lot of documentation on this; you can see the various values in the debugger; I also found amex using it: [snapshot_summary.rb](https://github.com/americanexpress/xcode-result-bundle-processor/blob/master/lib/xcoderesultbundleprocessor/snapshot_summary.rb)

mykola-mokhnach:
I checked the Ruby code and there it is verified (probably, because of a reason) whether an element provides the corresponding attribute first before getting the value.

```ruby
if element_snapshot.key?(:additionalAttributes) and element_snapshot[:additionalAttributes].key?(5004.to_s.to_sym)
        element_class = element_snapshot[:additionalAttributes][5004.to_s.to_sym]
      end
```

This means the property might not be consistent to all UI elements - some of them may have it and some other may not.

qmfrederik:
Hmm, a blame on that file shows that code is 2 years old and there's not much info in the logs.

@mykola-mokhnach What's your take on it? Worth adding, even if it may not be present for all elements?

mykola-mokhnach:
This property probably shows the old class name used by UIAutomation. If this is really like this then I saw some cases when UIAutomation class name was more specific than XCTest's one (which is always **Other** for "unknown" elements). Although I'm not sure about whether one can rely on it, since Apple didn't document it anywhere. Apple might remove the property one day after iOS update and then all automation clients will have to change their locators.

qmfrederik:
@marekcirkos @mykola-mokhnach I updated the PR so that the code which fetches the class, is now part of FBElement. The main purpose is to get some additional information which can be used for debugging.

For example, for the status bar you would get values such as:
```
<XCUIElementTypeOther type="XCUIElementTypeOther" name="Bluetooth on" class="UIStatusBarBluetoothItemView" label="Bluetooth on" enabled="true" visible="true" x="323" y="0" width="8" height="20"/>
```

i.e. the `class="UIStatusBarBluetoothItemView"` communicates some information about the element which you cannot get from the `XCUIElementTypeOther` element type.

Let me know if you want to proceed with this.

WebDriverAgentLib/Categories/XCUIElement+FBClass.m
```diff
@@ -0,0 +1,37 @@
...
+
+@end
+
+@implementation XCElementSnapshot (FBClass)
+
+- (NSString*)fb_class
+{
+  NSNumber* classId = [NSNumber numberWithInt:5004];
+  NSObject* class = [self.additionalAttributes objectForKey:classId];
+  if([class isKindOfClass:[NSString class]]){
```
mykola-mokhnach: check if **class** is not equal to nil first

WebDriverAgentLib/Categories/XCUIElement+FBClass.m
```diff
@@ -0,0 +1,37 @@
...
+    return (NSString*)class;
+  } else {
```
mykola-mokhnach: _else_ is not needed here

WebDriverAgentLib/Categories/XCUIElement+FBClass.m
```diff
@@ -0,0 +1,37 @@
...
+#import "XCUIElement+FBClass.h"
+
+@implementation XCUIElement (FBClass)
+
+- (NSString*)fb_class
+{
+  if (!self.lastSnapshot) {
+    [self resolve];
+  }
+  return [self.lastSnapshot fb_class];
```
mykola-mokhnach: there is fb_lastSnapshot method in Utilities category, which resolves the element automatically

WebDriverAgentLib/Categories/XCUIElement+FBClass.h
```diff
@@ -0,0 +1,30 @@
...
+
+#import <WebDriverAgentLib/XCElementSnapshot.h>
+#import <WebDriverAgentLib/XCUIElement.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface XCUIElement (FBClass)
+
+/*! Element's class (UIView or derived) */
+@property (nonatomic, readonly, copy) NSString *fb_class;
```
mykola-mokhnach: is it nullable or not?

WebDriverAgentLib/Categories/XCUIElement+FBWebDriverAttributes.m
```diff
@@ -90,6 +91,11 @@ - (NSString *)wdType
   return [FBElementTypeTransformer stringWithElementType:self.elementType];
 }
 
+- (NSString *)wdClass
+{
+  return FBTransferEmptyStringToNil(self.fb_class);
```
mykola-mokhnach: what the point to transform it to nil here if this can be already done in the corresponding extension method?

WebDriverAgentLib/Utilities/FBXPath.m
```diff
@@ -224,6 +224,13 @@ + (int)recordElementAttributes:(xmlTextWriterPtr)writer forElement:(XCElementSna
       return rc;
     }
   }
+  if (element.wdClass) {
```
mykola-mokhnach: this is going to break several unit and integration tests. Please check and update these

WebDriverAgentLib/Categories/XCUIElement+FBClass.m
mykola-mokhnach:
```objc
id class = [self.additionalAttributes objectForKey:classId];
if (class && [class isKindOfClass:[NSString class]] && [(NSString *)class length]) {
  return (NSString *)class;
}
return nil;
```

marekcirkos: If it is `nil` it will resolve to false anyway

WebDriverAgentLib/Categories/XCUIElement+FBClass.m
```diff
@@ -0,0 +1,37 @@
...
+@implementation XCElementSnapshot (FBClass)
+
+- (NSString*)fb_class
+{
+  NSNumber* classId = [NSNumber numberWithInt:5004];
+  NSObject* class = [self.additionalAttributes objectForKey:classId];
+  if([class isKindOfClass:[NSString class]]){
+    return (NSString*)class;
+  } else {
+    return @"";
```
marekcirkos: Why not use `nil` here?

WebDriverAgentLib/Categories/XCUIElement+FBClass.m
```diff
@@ -0,0 +1,37 @@
...
+}
+
+@end
+
+@implementation XCElementSnapshot (FBClass)
+
+- (NSString*)fb_class
+{
+  NSNumber* classId = [NSNumber numberWithInt:5004];
+  NSObject* class = [self.additionalAttributes objectForKey:classId];
```
marekcirkos: Please use `id` instead

WebDriverAgentLib/Categories/XCUIElement+FBClass.h
```diff
@@ -0,0 +1,30 @@
...
+
+@end
+
+
+@interface XCElementSnapshot (FBClass)
```
marekcirkos: Probably `FBUIClassName` would be better

marekcirkos:
I am bit resistan to this diff in general. What is the gain here? You are duplicating `XCT` class names here. Plus client shouldn't care much about classes anyway.

@mykola-mokhnach any thoughts on this?

mykola-mokhnach:
@marekcirkos this can sometimes help to locate an element if it has no accessibility identifier and XCTest recognises it as ElementTypeOther. This is the only possible advantage there. Also, I'm not sure about the actual performance. @qmfrederik do you see any visible perf degradation while getting the same source with and without class property?

mykola-mokhnach:
Altogether I tend to agree with @marekcirkos 
There is no point to add more attributes unless they are really necessary.

qmfrederik:
@mykola-mokhnach @marekcirkos 

Thanks for the feedback!

As to the "why": yes, the idea is to get additional information for elements that come back as "XCUIElementTypeOther", for example in the status bar you have:

```
<XCUIElementTypeOther type="XCUIElementTypeOther" class="UIStatusBarBluetoothItemView" />
```

So it's not merely duplicating; it's adding additional information.

As to the performance: I measured it; it's essentially reading a value from a dictionary (and that dict was prepopulated); the cost (on an iPhone 7) is in the range of nanoseconds per call; so even if you have say 100 elements it will only add 1 ms to the output.

As to unit tests: certainly, I'll add those!

marekcirkos:
@qmfrederik I really appreciate your support here. However I can't see the point of serving this information. 
- You wouldn't be able to search elements for those classes or it will be really slow as it will need to serialise every element. Main reason why xpath is slow. Probably you want to use `xpath` anyway?
- It will be confusing which class name should be used to locate element with `by class name` search.

qmfrederik:
@marekcirkos The main use case here is people who are trying to "dynamically" automate apps - that is, they have certain algorithms, such as "try to log on" which will try to find the user name + password text boxes and the login button and then enter credentials and try to log in.

In those cases, they heavily rely on /source (hence also the other PR related to visible) to get a full dump of the UI hierarchy, execute an algorithm to identify the relevant elements, and then execute actions against it.

So basically their requirement is to have as much information in /source as possible. It's fine if not all that information can be used in XPath or other queries - they can work around it.

I'm also OK with you not taking this PR - we try to contribute back our modifications to upstream open source projects, but it's up to you to define what you want to take and what not :)

marekcirkos:
I wonder if it would make sense to drop XCTest class names in favour of iOS class names.

@mykola-mokhnach any thoughts on this?

mykola-mokhnach:
And break thousands of existing automation scripts? No thanks :)

marekcirkos:
I am sorry we will have to pass on this.

PR #591, opened 2017-05-18
Make /source about 3x faster (by making visible less accurate)

qmfrederik:
The `/source` route is very slow; on older devices (iPhone 5S), I get response times of approx 3.7s for retrieving the source of a home page; on newer devices (iPhone 7), the response times are about 1s but still slow.

Most of the time is spent in:
- Getting the active application
- Determining the visibility of the individual elements

I didn't dig into fetching the application yet, but when determining the visibility, the visible rectangle and hitpoint are determined for each element with a non-zero size and this is a *very* expensive operation, especially in a tight loop.

This PR adds a "heuristic" flag to isVisible; if this flag is set:
- `visibleFrame` is assumed to be equal to `frame`
- the hit point is always assumed to be inside the application frame.

This flag is *only* used inside the `/source` route, not in other routes.

The reasoning is that when you want to get the `/source`, you are usually a bit less strict about the accuracy of the visibility and can always call isVisible for elements about which you care. Other options exist as well - we could make it a configuration flag, add a toggle to the route or create a newer route.

Here are the response times for the `/source` route before and after, on iOS 10:

| Device         | Before                  | After
|------------|-------------------|-------
| iPhone 5S  | 3770ms                  | 850ms
| iPhone 7     | 1000ms                  | 350ms

In the after time, most time is spend getting the active application, we still may be able to optimize that further.

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.h
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.h
@@ -25,6 +25,9 @@ NS_ASSUME_NONNULL_BEGIN
 /*! Whether or not the element is visible */
 @property (atomic, readonly) BOOL fb_isVisible;
 
+/*! Whether or not the element is visible, optionally using a 
+ *  less precise heuristic o bypass a very expensive hittest */
+- (BOOL)fb_isVisible:(BOOL)useHeuristic;
 @end
 
 NS_ASSUME_NONNULL_END
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
@@ -30,10 +30,19 @@
 @end
 
 @implementation XCElementSnapshot (FBIsVisible)
-
 - (BOOL)fb_isVisible
 {
-  if (CGRectIsEmpty(self.frame) || CGRectIsEmpty(self.visibleFrame)) {
+  return [self fb_isVisible:NO];
+}
+
+- (BOOL)fb_isVisible: (BOOL)useHeuristic
+{
+  CGRect frame = self.frame;
+  if (CGRectIsEmpty(frame)) {
+    return NO;
+  }
+  CGRect visibleFrame = useHeuristic ? frame : self.visibleFrame;
+  if (CGRectIsEmpty(visibleFrame)) {
     return NO;
   }
   if ([FBConfiguration shouldUseTestManagerForVisibilityDetection]) {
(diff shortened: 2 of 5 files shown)


qmfrederik:
Again - putting this PR out there initially to gather feedback, let me know what your thoughts are and if/how we can land this.

mykola-mokhnach:
The change means the actual isDisplayed property of an element and the corresponding property in the /source output might be different. Is this really what we want?

qmfrederik:
@mykola-mokhnach Not really, but the performance of the `/source` route is a problem for us and something we'd like to address.

Other options include
1 Removing `isVisible` from `/source` alltogether
2 Use something else, like `isHidden` in `source` to indicate whether an element is explicitly marked as hidden (and exclude the hitpoint test); at least one WDA variant does it that way
3 Using stuff like `/source?heuristic=true` to indicate whether you want this behavior or not

(2) may be the better option here, what do you think?

... and others.

I'm wondering, on the Appium side of things, do you get feedback on the `/source` performance? Is it something you consider worth improving?

mykola-mokhnach:
> I'm wondering, on the Appium side of things, do you get feedback on the /source performance? Is it something you consider worth improving?

Usually /source call is used for debug purposes and "precision" there is more important than performance. 

Have you also tried to measure method performance if **shouldUseTestManagerForVisibilityDetection** property is set to YES?

mykola-mokhnach:
@marekcirkos What do you think?

qmfrederik:
> Have you also tried to measure method performance if shouldUseTestManagerForVisibilityDetection property is set to YES?

Yes, if this value is set to YES the performance is similar or worse, so > 4 sec on iPhone 5S and ~ 1 sec on iPhone 7

marekcirkos:
I am sorry, but serving bad information is a bad idea. 
We should either focus in improving it's performance in some other way, but keeping correctness
or add param to `/source` that will not return visibility param (if you really really really want this).

marekcirkos:
However it would be much better trying to understand why it is slow, and trying to resolve it. It will improve over whole WDA performance.

qmfrederik:
@marekcirkos Sure, I understand the concerns about `visible` not being accurate, so let's try to figure out what else we can do. #597 already looks like a good start (I'll try to get some perf data on it).

I measured the "slow" parts in the code by basically calculating how long each block of code takes in code itself; I did not manage to get a performance profiler to attach to WDA.

I tried running a performance trace on WDA but as soon as you launch Instruments in Xcode, the XCUITest session is interrupted - did you manage to get instruments to do performance profiling of WDA tests?

marekcirkos:
Closing as per discussion.

I don't think you can profile WDA with Xcode as it is a test. 
You can measure performance by calculating time spent in the code with `NSTimeIntervals`

PR #597, opened 2017-05-22
Do not try to calculate hitpoint for elements outside of the screen rect

mykola-mokhnach:
This is one of these cases where everything good should be simple. 
**self.fb_hitPoint** call is expensive and there is no point to call it if we already know that rectIntersects verification failed

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
@@ -33,18 +33,23 @@
 
 - (BOOL)fb_isVisible
 {
-  if (CGRectIsEmpty(self.frame) || CGRectIsEmpty(self.visibleFrame)) {
+  if (CGRectIsEmpty(self.frame)) {
+    return NO;
+  }
+  CGRect visibleFrame = self.visibleFrame;
+  if (CGRectIsEmpty(visibleFrame)) {
     return NO;
   }
   if ([FBConfiguration shouldUseTestManagerForVisibilityDetection]) {
     return [(NSNumber *)[self fb_attributeValue:FB_XCAXAIsVisibleAttribute] boolValue];
   }
-  XCElementSnapshot *app = [self _rootElement];
-  CGSize screenSize = FBAdjustDimensionsForApplication(app.frame.size, (UIInterfaceOrientation)[XCUIDevice sharedDevice].orientation);
+  CGRect appFrame = [self _rootElement].frame;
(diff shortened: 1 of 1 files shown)


marekcirkos:
Good catch!

qmfrederik:
@mykola-mokhnach Did you run perf tests on this? Here's what I get (using 10 requests):

* Current master: 781 ms +/- 10ms
* PR 597: 763 ms +/- 14ms

So about a 3% improvement in performance.

The gist of this is that `fb_hitPoint` is expensive, but `visibleFrame` is expensive, too. Both seem to re-use some cached variables; so if you call `visibleFrame` first, `fb_hitPoint` is fast(er).

Looking at the code, I was wondering if the following is true:

> If `self.frame` is entirely within `app.frame`, then `self.visibleFrame` and `self.fb_hitPoint` are also in `app.frame`

and

> If `self.frame` is entirely outside of `app.frame`, then `self.VisibleFrame` and `self.fb_hitPoint` are also outside of `app.Frame`

If that is true, you can avoid calling `visibleFrame` and `fb_hitPoint` in a lot of scenarios, and then I get the following results: 

* 552 ms +/- 7 ms

so about 30% improvement in performance.

mykola-mokhnach:
thanks for the hint @qmfrederik 
I have replaced **visibilityFrame** property with **frame** in the condition. It's clear that if element's **frame** is outside of the visible viewport then **visibleFrame** will be outside of it as well.
Can you please the perf now?

qmfrederik:
@mykola-mokhnach I re-ran the tests, but the results are pretty much the same. I think there is already some caching involved so it's the first call to hitpoint / visible frames which really cause the perf hit.

I think the fix would be to figure out if there's another way to anticipate the hitpoint will be either valid (inside the screen area) or invalid ( such as -1,-1). I'm guessing there must be another property which can give this information, but I didn't find one.

I'll probably keep looking at this the coming days and keep you updated.

PS:
- To measure the HTTP response time, I just load the /source URL with Chrome and use the Chrome developer tools
- To figure out which code blocks are taking time, I use `CFTimeInterval`:

```
(...)

#import <QuartzCore/QuartzCore.h>

(...)

  CFTimeInterval startTime = CACurrentMediaTime();

  if (CGRectIsEmpty(self.frame)) {
    return NO;
  }
  CFTimeInterval evaluateFrameEmptyTime = CACurrentMediaTime();

(...)


  NSLog(@"frameEmpty: %f, visibleFrame: %f, configuration: %f, frameIntersect: %f, hitPoint: %f, visible: %d, x: %f, y: %f, width: %f, height: %f, hitPoint: (%f, %f)",
        evaluateFrameEmptyTime - startTime,

(...)

```

PR #598, opened 2017-05-23
Avoid xpath usage for alert text lookup

mykola-mokhnach:
The implementation of fb_descendantsMatchingType is slow and implicitly uses slower xpath lookup (which sometimes causes unexpected 'Copy attribute error' 15s delays). It would be nice to get rid of _fb_descendantsMatchingType_ and _fb_descendantsMatchingXPathQuery_ extension methods from **XCElementSnapshot+FBHelpers** category and I'd like to take care about it after https://github.com/facebook/WebDriverAgent/pull/569 is merged

### DIFF
--- a/WebDriverAgentLib/FBAlert.m
+++ b/WebDriverAgentLib/FBAlert.m
@@ -102,7 +102,8 @@ NSString *const FBAlertObstructingElementException = @"FBAlertObstructingElement
   if (resultText.count) {
     return [resultText componentsJoinedByString:@"\n"];
   }
-  return nil;
+  // return null to reflect the fact there is an alert, but it does not contain any text
+  return (id)[NSNull null];
 }
 
 - (NSArray *)buttonLabels
--- a/WebDriverAgentLib/FBAlert.m
+++ b/WebDriverAgentLib/FBAlert.m
@@ -92,9 +92,9 @@ NSString *const FBAlertObstructingElementException = @"FBAlertObstructingElement
   if (!alert) {
     return nil;
   }
-  NSArray<XCElementSnapshot *> *staticTextList = [alert.lastSnapshot fb_descendantsMatchingType:XCUIElementTypeStaticText];
+  NSArray<XCUIElement *> *staticTextList = [alert descendantsMatchingType:XCUIElementTypeStaticText].allElementsBoundByIndex;
   NSMutableString *mText = [NSMutableString string];
-  for (XCElementSnapshot *staticText in staticTextList) {
+  for (XCUIElement *staticText in staticTextList) {
     if (staticText.wdLabel && staticText.isWDVisible) {
       [mText appendFormat:@"%@\n", staticText.wdLabel];
     }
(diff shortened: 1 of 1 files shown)


WebDriverAgentLib/FBAlert.m
```diff
@@ -92,18 +92,17 @@ - (NSString *)text
...
     }
   }
-  // Removing last '\n'
-  if (mText.length > 0) {
-    [mText replaceCharactersInRange:NSMakeRange(mText.length - @"\n".length, @"\n".length) withString:@""];
+  if (resultText.count) {
+    return [resultText componentsJoinedByString:@"\n"];
   }
-  return mText.length > 0 ? mText.copy : [NSNull null];
+  return nil;
```
marekcirkos: Please return `[NSNull null]` instead.

WebDriverAgentLib/FBAlert.m
```diff
@@ -92,18 +92,17 @@ - (NSString *)text
...
   }
-  NSArray<XCElementSnapshot *> *staticTextList = [alert.lastSnapshot fb_descendantsMatchingType:XCUIElementTypeStaticText];
-  NSMutableString *mText = [NSMutableString string];
-  for (XCElementSnapshot *staticText in staticTextList) {
+  NSArray<XCUIElement *> *staticTextList = [alert descendantsMatchingType:XCUIElementTypeStaticText].allElementsBoundByIndex;
+  NSMutableArray<NSString *> *resultText = [NSMutableArray array];
+  for (XCUIElement *staticText in staticTextList) {
     if (staticText.wdLabel && staticText.isWDVisible) {
-      [mText appendFormat:@"%@\n", staticText.wdLabel];
+      [resultText addObject:[NSString stringWithFormat:@"%@", staticText.wdLabel]];
```
marekcirkos: Is `[NSString stringWithFormat:` even needed here? Why not simply `staticText.wdLabel`?

mykola-mokhnach: because wdLabel is nullable, which means I'll have to perform types conversation anyway

WebDriverAgentLib/FBAlert.m
mykola-mokhnach: maybe I do something wrong, but the compiler didn't like it

marekcirkos: Probably class mismatch.

WebDriverAgentLib/FBAlert.m
marekcirkos: I see. You could do shorter `staticText.wdLabel ?: @""`, however `stringWithFormat` is fine in that case as well.

WebDriverAgentLib/FBAlert.m
mykola-mokhnach: Yep, it complains about incompatible pointer types (NSString* _Nullable vs NSNull* _Nonnull)
Any suggestions how to workaround it properly?

marekcirkos: We can either:
1) Force cast it to `id` and add comment
2) return `nil` and handle it when reporting alert text.

marekcirkos: Second you would be preferable, but probably requires a bit more work.

mykola-mokhnach: added a comment and forced typecast

Issue #586, opened 2017-05-17
scrolling with an iOS predicate using name doesn't work


oscheller1:
## Description:
Running the command in the Python Appium client:
```python
self.driver.execute_script('mobile: scroll', {"element": scrollable_parent, "predicateString": "name == 'settings_debugging_option_validation'"})
```
throws the following error (see line 4):

```
Calling AppiumDriver.execute() with args: ["mobile: scroll",[{"predicateString":"name == 'settings_debugging_option_validation'","element":{"ELEMENT":"1E83769D-5121-402E-900E-C803B07D79CC","element-6066-11e4-a52e-4f735466cecf":"1E83769D-5121-402E-900E-C803B07D79CC"}}],"d0b46a3a-66de-46ae-98c9-c91cbd314c46"]
[debug] [XCUITest] Executing command 'execute'
[debug] [JSONWP Proxy] Proxying [POST /wda/element/1E83769D-5121-402E-900E-C803B07D79CC/scroll] to [POST http://localhost:8100/session/BFC1CD73-11EF-4A86-81EA-0C7B9292AD36/wda/element/1E83769D-5121-402E-900E-C803B07D79CC/scroll] with body: {"predicateString":"name == 'settings_debugging_option_validation'"}
[debug] [JSONWP Proxy] Got response with status 200: {"value":"[<XCElementSnapshot 0x7ffad6a18130> valueForUndefinedKey:]: this class is not key value coding-compliant for the key name.\n\n(\n\t0   CoreFoundation                      0x0000000107a6dd85 __exceptionPreprocess + 165\n\t1   libobjc.A.dylib                     0x00000001074e1deb objc_exception_throw + 48\n\t2   CoreFoundation                      0x0000000107a6d9c9 -[NSException raise] + 9\n\t3   Foundation                          0x000000010713f144 -[NSObject(NSKeyValueCoding) valueForUndefinedKey:] + 226\n\t4   Foundation                          0x0000000107095167 -[NSObject(NSKeyValueCoding) valueForKey:] + 280\n\t5   Foundation                          0x00000001070cfa55 -[NSFunctionExpression expressionValueWithObject:context:] + 1096\n\t6   Foundation                          0x00000001070cf532 -[NSComparisonPredicate evaluateWithObject:substitutionVariables:] + 243\n\t7   Foundation                          0x00000001070cf3d0 _filterObjectsUsingPredicate + 398\n\t8   Foundation          ...
[...]

PR #604, opened 2017-05-31
Fix https://github.com/facebook/WebDriverAgent/issues/586

mykola-mokhnach:
This PR fixes two potential issues in **scroll** API:
- unexpected errors in case no UI elements were matched by given predicate/identifier
- key error while passing unformatted predicate expression

### DIFF
--- a/WebDriverAgentLib/Commands/FBElementCommands.m
+++ b/WebDriverAgentLib/Commands/FBElementCommands.m
@@ -20,6 +20,7 @@
 #import "FBApplication.h"
 #import "FBMacros.h"
 #import "FBMathUtils.h"
+#import "NSPredicate+FBFormat.h"
 #import "XCUICoordinate.h"
 #import "XCUIDevice.h"
 #import "XCUIElement+FBIsVisible.h"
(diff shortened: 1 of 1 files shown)


WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -243,8 +244,10 @@ + (NSArray *)routes
   // what ios-driver did and sadly, we must copy them.
   NSString *const name = request.arguments[@"name"];
   if (name) {
-    NSPredicate *predicate = [NSPredicate predicateWithFormat:@"%K == %@", FBStringify(XCUIElement, wdName), name];
-    XCUIElement *childElement = [[[[element descendantsMatchingType:XCUIElementTypeAny] matchingPredicate:predicate] allElementsBoundByIndex] lastObject];
+    XCUIElement *childElement = [[[[element descendantsMatchingType:XCUIElementTypeAny] matchingIdentifier:name] allElementsBoundByIndex] lastObject];
```
marekcirkos: Why not `wdName`?

mykola-mokhnach: I assume search by identifier (which is the same values as in _wdName_) works faster and also we managed to get rid of one extra code line 💯

PR #617, opened 2017-06-12
Add self element to descendants lookup

mykola-mokhnach:
This PR adds self element to the list of possible descendants matches.

The problem has been discovered while creating tests for https://github.com/facebook/WebDriverAgent/pull/612

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBUtilities.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBUtilities.m
@@ -75,6 +75,9 @@ static const NSTimeInterval FBANIMATION_TIMEOUT = 5.0;
   NSArray<NSNumber *> *matchedUids = [snapshots valueForKey:FBStringify(XCUIElement, wdUID)];
   NSMutableArray<XCUIElement *> *matchedElements = [NSMutableArray array];
   if ([matchedUids containsObject:@(self.wdUID)]) {
+    if (1 == snapshots.count) {
+      return @[self];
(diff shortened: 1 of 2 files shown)


WebDriverAgentLib/Categories/XCUIElement+FBUtilities.m
```diff
@@ -73,29 +73,35 @@ - (XCElementSnapshot *)fb_lastSnapshot
     return @[];
   }
   NSArray<NSNumber *> *matchedUids = [snapshots valueForKey:FBStringify(XCUIElement, wdUID)];
+  NSMutableArray<XCUIElement *> *matchedElements = [NSMutableArray array];
+  if ([matchedUids containsObject:@(self.wdUID)]) {
+    if (1 == snapshots.count) {
+      return @[self];
```
marekcirkos: Why you assume self matches the request here?

mykola-mokhnach: _snapshots_ array defines how many elements should be returned. The situation where single snapshot matched to multiple elements is not possible. This means if we only need to match one snapshot and the uid of this snapshot matched to the uid of self element then the result should be single self element

marekcirkos: Can you add test case for this?

mykola-mokhnach: done

Issue #615, opened 2017-06-12
Screen rotation is reset if an alert is visible


alexander-poulikakos:
I don't have the access to re-open issue https://github.com/facebook/WebDriverAgent/issues/300. So opening this instead. 

This is still a problem. See https://github.com/facebook/WebDriverAgent/issues/300 for full details.

marekcirkos:
Re-opend https://github.com/facebook/WebDriverAgent/issues/300

PR #623, opened 2017-06-15
Add a workaround for XCTest bug with landscape tap for application restore

mykola-mokhnach:
We got several complains in Appium about application restore does not work properly for landscape mode. I investigated the issue and it looks like landscape tap works as expected starting from iOS 10.0, but only for system apps (and springboard is one of these). 
Also added a separate integration test for iPad only to confirm that.

This PR should also add a workaround for clicking alert buttons described in the issue https://github.com/facebook/WebDriverAgent/issues/615

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBTap.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBTap.m
@@ -16,6 +16,7 @@
 #import "XCUIElement+FBUtilities.h"
 #import "XCEventGenerator.h"
 #import "XCSynthesizedEventRecord.h"
+#import "XCElementSnapshot+FBHelpers.h"
 #import "XCElementSnapshot+FBHitPoint.h"
 #import "XCPointerEventPath.h"
 #import "XCTRunnerDaemonSession.h"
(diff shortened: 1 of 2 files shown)


WebDriverAgentLib/Categories/XCUIElement+FBTap.m
```diff
@@ -26,13 +27,35 @@ @implementation XCUIElement (FBTap)
 
 - (BOOL)fb_tapWithError:(NSError **)error
 {
+  if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(@"10.0") && [self fb_hasInvertedParentWindow]) {
+    // This indicates the fact, that the element is not affected by XCTest landscape bug and
+    // is properly clicked by calling the native tap selector
+    [self tap];
```
marekcirkos: The whole reason of having `fb_tapWithError` is to detect tap failures which are not reported by `tap`. Couldn't you simply alter hitpoint calculation here?

mykola-mokhnach: I tried, but this is real black magic to me how it works internally and in which way these bloody coordinates are corrupted by XCTest internals. After spending hours on investigation I decided to stay with this approach, because it's proven to work.

marekcirkos: I'll grab your test and have a look.

marekcirkos:
Test you added is passing for me on current build.

mykola-mokhnach:
@marekcirkos Do you try Xcode9 or Xcode8?

mykola-mokhnach:
Yep, it magically started to work properly now even without this patch. Closing.

PR #630, opened 2017-06-27
Add basic support for tvOS

georgmay:
Hello!

Unfortunately, this pull request become massive due to adding tvOS private headers. 
In general, this PR should add basic tvOS support to make it possible to launch webdriver under tvOS.

There are some changes done due to differences between iOS & tvOS.
1. Different base SDKs for tvOS and iOS, weird behaviour of headermap which I was trying to play with to make it possible ignoring project headers, led me to create separate project pointing to the same codebase, but having different config in build settings, i.e. base sdk to be *tvOS*, specific plist file for tvOS integration app etc.
2. *XCUICoordinate* physically doesn't exist for tvOS, which make sense, since there's no restriction-free pointing device for tvOS, hence its using has been limited for non-tvos code only.
3. Since there are different storyboard types, I've added tvOS storyboard for integration app and linked it in *Info-tvOS.plist* file.

***Documentation update***:
I've also added a couple of new endpoint to support apple remote commands, will update documentation once we agree that this is a way to go. 
Also, I would add the commands to launch it for tvOS simulator and devices.

***Dependency update***:
I have updated ***RoutingHTTPServer*** in my fork to support tvOS, although, since this library has become outdated, there is probably a good sign to replace it with something newer, however it should be a part of different PR.
I can create a pull request into https://github.com/marekcirkos/RoutingHTTPServer to add tvOS support.

There is a tricky navigation issue as well, although since this PR is already massive, I decided to split them up.

I am happy to hear any feedback and fix any issues! Thanks for your time!


marekcirkos:
Thank you for your work! If you wouldn't mind, I would like to resolve Xcode 9 changes and than extend tvOS support. Otherwise we will be fixing to many things at the same time.

georgmay:
Thanks for reviewing! 
That's fine, although I can't run CI properly because, I guess, the github token is being used to fetch this fork https://github.com/marekcirkos/RoutingHTTPServer, so I guess it'd better to open a pull request into this fork?

marekcirkos:
I am happy to change dependency. Do you have any suggestion?

marekcirkos:
Also I'd rather make this migration in smaller steps. e.g.
1) Add headers
2) Add targets.
3) Fixing endpoints etc.

I'll close this PR, because you will have million of conflicts anyway and created Issue (https://github.com/facebook/WebDriverAgent/issues/640) to discuss it, if you are still interested :)

PR #641, opened 2017-07-10
Implement firstMatch call support for iOS 11+

mykola-mokhnach:
[firstMatch](https://developer.apple.com/documentation/xctest/xcuielementtypequeryprovider/2902250-firstmatch?language=objc) implements fast short-circuit traversal of UI tree, which means better lookup times for us.

I've also changed some existing methods to use this call and make them faster.

Tested on Xcode8 with iOS 10 and Xcode 9b2 + iOS 11beta.

I have noticed one new issue under iOS11 - the _children_ property of XCElementSnapshot returns empty list without any known reason. The same code works as expected under iOS 10.

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBUtilities.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBUtilities.m
@@ -15,6 +15,7 @@
 #import "FBMathUtils.h"
 #import "FBPredicate.h"
 #import "FBRunLoopSpinner.h"
+#import "FBXCodeCompatibility.h"
 #import "XCAXClient_iOS.h"
 #import "XCUIElement+FBWebDriverAttributes.h"
 
@@ -84,8 +85,13 @@ static const NSTimeInterval FBANIMATION_TIMEOUT = 5.0;
   if (uniqueTypes && [uniqueTypes count] == 1) {
     type = [uniqueTypes.firstObject intValue];
   }
-  [matchedElements addObjectsFromArray:[[self descendantsMatchingType:type] matchingPredicate:[FBPredicate predicateWithFormat:@"%K IN %@", FBStringify(XCUIElement, wdUID), matchedUids]].allElementsBoundByIndex];
-  if (matchedElements.count <= 1) {
+  XCUIElementQuery *query = [[self descendantsMatchingType:type] matchingPredicate:[FBPredicate predicateWithFormat:@"%K IN %@", FBStringify(XCUIElement, wdUID), matchedUids]];
+  if (1 == snapshots.count) {
+    XCUIElement *result = query.fb_firstMatch;
+    return result ? @[result] : @[];
+  }
+  NSArray<XCUIElement *> *filteredElements = query.allElementsBoundByIndex;
+  if (filteredElements.count <= 1) {
     // There is no need to sort elements if count of matches is not greater than one
     return matchedElements;
   }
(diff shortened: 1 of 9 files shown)


marekcirkos:
Should we wait with this for full Xcode 9 compatibility?

mykola-mokhnach:
I personally don't see any reason to wait it will use the new approach only if it is available. And the old approach still works.

marekcirkos:
Can you resolve conflicts after https://github.com/facebook/WebDriverAgent/pull/612 lands?

mykola-mokhnach:
Sure thing. I also have a TODO in that PR  to add firstMatch support

mykola-mokhnach:
@marekcirkos conflict resolved

marekcirkos:
Do you see performance improvements, if you rebase on master? I wander if it makes sense to land this or rather kill snapshots first?

mykola-mokhnach:
@marekcirkos I don't observe any visible difference in lookup performance. However I would vote for merging this, because it is for sure not slower than the previous implementation.

WebDriverAgentLib/Categories/XCUIElement+FBUtilities.m
```diff
@@ -84,10 +85,15 @@ - (XCElementSnapshot *)fb_lastSnapshot
   if (uniqueTypes && [uniqueTypes count] == 1) {
     type = [uniqueTypes.firstObject intValue];
   }
-  [matchedElements addObjectsFromArray:[[self descendantsMatchingType:type] matchingPredicate:[FBPredicate predicateWithFormat:@"%K IN %@", FBStringify(XCUIElement, wdUID), matchedUids]].allElementsBoundByIndex];
+  XCUIElementQuery *query = [[self descendantsMatchingType:type] matchingPredicate:[FBPredicate predicateWithFormat:@"%K IN %@", FBStringify(XCUIElement, wdUID), matchedUids]];
+  if (1 == snapshots.count) {
```
marekcirkos: I think this will resolve whole query anyway, so you have no benefit of using it here. Unless count is optimised here.

WebDriverAgentLib/Utilities/FBXCodeCompatibility.m
```diff
@@ -43,3 +43,28 @@ + (instancetype)fb_applicationWithPID:(pid_t)processID
 
 @end
 
+static BOOL FBShouldUseFirstMatchSelector = NO;
+static dispatch_once_t onceFirstMatchToken;
+@implementation XCUIElementQuery (FBCompatibility)
+
+- (XCUIElement *)fb_firstMatch
+{
+  SEL firstMatchSelector = NSSelectorFromString(@"firstMatch");
```
marekcirkos: Possibly, you could simply add firstMatch selector in our private headers as class extension (`@interface XCUIEment ()`) and keep  `respondsToSelector` check. That would be even faster

mykola-mokhnach: originally firstMatch is a part of XCUIElementTypeQueryProvider protocol. Shall I add it there or it's better to do it directly in XCUIElement?

WebDriverAgentLib/Categories/XCUIElement+FBUtilities.m
mykola-mokhnach: **snapshots** is the simple array, that we receive as method argument. No resolution is happening here.

WebDriverAgentLib/Utilities/FBXCodeCompatibility.m
mykola-mokhnach: however XCTest headers prior to XCode9 don't contain such protocol and only have XCUIElementQuery

nghiadhd:
I got weird issue, I tried to send 2 press home button events to show AppSwitcher (Recent Apps)
From iOS 9.3 -> iOS 10.2.x -> OK
From iOS 10.3 -> iOS 11 Beta 4 -> Not work

I tested on iPhone & iPad. 
Do any body know how to workaround?

PR #642, opened 2017-07-10
Removed old visibility check

marekcirkos:
New visibility check seems to be good enough, so removing old one.

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
@@ -39,9 +39,6 @@
   if (CGRectIsEmpty(visibleFrame)) {
     return NO;
   }
-  if ([FBConfiguration shouldUseTestManagerForVisibilityDetection]) {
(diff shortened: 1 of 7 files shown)


marekcirkos:
@mykola-mokhnach Do you guys still care about this or you are happy to remove it?

mykola-mokhnach:
We have a special capability to use the alternative visibility detection method in Appium, but I have only seen 2 or 3 situations where we advised to use it. @imurchie what do you think, do we still need it?

imurchie:
iOS 9.3 does not work at all using the new method, so yes.

marekcirkos:
@imurchie It doesn't work because it is wrong or it crashes?

marekcirkos:
https://github.com/imurchie?

imurchie:
It always returns "true", if memory serves me correctly. Scrolling and stuff doesn't work because all elements are considered to already be on screen.

marekcirkos:
Eh, that's a shame. Ok I'll drop it for now, so that we don't cause you too much trouble. :)

imurchie:
Our plan is to drop iOS 9 support once iOS 11 is released for real. On our end there is a tonne that is needed for iOS 9 support (including supporting Instruments) so we are happy about this.

PR #648, opened 2017-07-13
Fix getting element UID in Xcode 9

mykola-mokhnach:
In the recent XCTest release instances of XCAccessibilityElement don't contain the **payload** property anymore. And the unique element's identifier has been moved to _elementID property.


marekcirkos:
Awesome, I was about to work on this today :)

Issue #639, opened 2017-07-10
Xcode 9 Support


marekcirkos:
Time to start working on Xcode 9 :). Please raise any issues here. I'll keep updating todo list to reflect known, left issues.

- [x] WDA is not building

- [x]  `_parentElement` selector is missing

- [x]  `appWithPID:` selector is missing

- [x] `rect` request fails to wrap NSNumber
- [x] testSimpleScroll - Scrolling seems to work, but this test fails what seems to be change on how touch events are handled

- [x] XCUIDeviceRotationTests - Rotation works, but `orientation` property is not being updated,

- [x] testInvisibleDescendantWithXPathQuery - searched element is now considered visible.

- [ ] `source` is slow

- [x] performance is much worse

- [x] WebView subviews are not visible

imurchie:
When trying to do anything involving numbers JSON parsing fails. For instance, getting the `rect` (this is from an Appium log, but you can see the URL we are hitting in WebDriverAgent, and the error we get back):
```
dbug JSONWP Proxy Proxying [GET /element/A6FB5912-4DC1-4CC3-A125-8913F3854717/rect] to [GET http://localhost:8100/session/B346E0B7-E439-43DD-B7AF-8DA6F32691BA/element/A6FB5912-4DC1-4CC3-A125-8913F3854717/rect] with no body
[...]

PR #649, opened 2017-07-14
Use application.frame property to make visibility property work for iOS 11

mykola-mokhnach:
Without this patch XCode 9 generates the stracktrace mentioned in https://github.com/facebook/WebDriverAgent/issues/639 for each call of fb_isVisible.

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
@@ -42,7 +42,7 @@
   if ([FBConfiguration shouldUseTestManagerForVisibilityDetection]) {
     return [(NSNumber *)[self fb_attributeValue:FB_XCAXAIsVisibleAttribute] boolValue];
   }
-  CGRect appFrame = [self fb_rootElement].frame;
+  CGRect appFrame = self.application.frame;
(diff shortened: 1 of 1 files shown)


marekcirkos:
Does it work? I think I tried it and it didn't work. Let me check again.

mykola-mokhnach:
I tried Xcode 9b2 + iOS11 on iPhone 7 simulator and this worked like a charm

marekcirkos:
There is no harm in landing it anyway. :)

marekcirkos:
Yeah, it didn't help :/

mykola-mokhnach:
@marekcirkos Interesting. Which exception do you get?

marekcirkos:
I didn't get any exception, even without your change. However sometimes this returns self instead of application snapshot. I got a fix for this, but it is nasty and I am trying to find other solution.

Dheepthi:
Is this fixed? 
I get the below exception when trying to build the latest WDA:
Commit id : f4fab7df9e42673f39fe610082f45c928b646245
Title: Use visibility of children rather than hitpoint of descendants.

WebDriverAgent-master/WebDriverAgentLib/Routing/FBSession.m:77:10: Implicit conversion from nullable pointer 'FBApplication * _Nullable' to non-nullable pointer type 'FBApplication * _Nonnull'

PR #650, opened 2017-07-14
Use "official" visibility detection scheme since iOS 11 by default

mykola-mokhnach:
I think it might be a good idea to stick to the visibility detection method, which is "officially" provided by Apple since we anyway plan to deprecate most of the stuff related to snapshots since iOS 11. Also, this way it will be faster and more reliable.

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
@@ -9,6 +9,7 @@
 
 #import "XCUIElement+FBIsVisible.h"
 
+#import "FBMacros.h"
 #import "FBApplication.h"
 #import "FBConfiguration.h"
 #import "FBMathUtils.h"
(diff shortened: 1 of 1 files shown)


WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
```diff
@@ -23,7 +24,10 @@ @implementation XCUIElement (FBIsVisible)
 
 - (BOOL)fb_isVisible
 {
-  return self.fb_lastSnapshot.fb_isVisible;
+  if (SYSTEM_VERSION_LESS_THAN(@"11.0") || [FBConfiguration shouldUseTestManagerForVisibilityDetection]) {
```
marekcirkos: Can you remove `SYSTEM_VERSION_LESS_THAN` check? If that work it should work current iOS version and pass green on current tests.

mykola-mokhnach: Unlikely... at least on iOS9 snapshot's hitPoint property and hittable property of XCElement return different resulsts

marekcirkos: I fixed bunch of stuff. I don't think it is needed anymore. Is it?

PR #652, opened 2017-07-14
Check alert wdValue if wdLabel is empty

imurchie:
In iOS 11 the `wdLabel` on alerts always seems to be blank. `wdValue` has the same info as used to be in `wdLabel`, so try that out.


mykola-mokhnach:
Probably, this issue will be fixed after https://github.com/facebook/WebDriverAgent/pull/658 is merged

marekcirkos:

imurchie:
This seems to be fixed now (as does text retrieval in general!). Closing.

PR #653, opened 2017-07-15
Xpath lookup resurrection for iOS 11+

mykola-mokhnach:
The actual content of element attributes in iOS 11+ is resolved during search query execution at runtime. This means, that by executing descendants lookup by type ANY these descendants will be returned with all their properties set to default values unless one reads the corresponding XCUIElement's property from XCUIElementAttributes protocol to "fill" the necessary fields. XCTest only keep pointers to the located elements and will lazily resolve the corresponding attribute values upon their invocation.
For XML tree building this means we not only need to gather all descendants, but also read the corresponding XCUIElement (and not XCElementSnapshot) attributes at least once (however each XCUIElementAttributes's property call is not cached and invokes a new internal XCTest request).

The current implementation for iOS 11+ is much slower, than it is for iOS 10 and earlier versions, but it is reliable and passes unit and integration tests. All comments and proposals are welcome.


mykola-mokhnach:
I'm going to fix the lookup by predicate in iOS 11 after this PR is merged (the problem about SIGABRT described in https://github.com/facebook/WebDriverAgent/issues/639), but it requires the changes  that I did here to XCUIElement+FBWebDriverAttributes.m.

marekcirkos:
This is way to much changes at one go. You killed `forwardingTargetForSelector` and much more. I can't really reason on what is actually changing here. Can you split this and explain what is going on here?

mykola-mokhnach:
@marekcirkos Unfortunately I cannot split this, since all these changes are mandatory to make the feature work properly. I will be happy to explain everything in a chat, for example via Appium Slack or Google Hangouts.

mykola-mokhnach:
However, it is theoretically possible to keep WebDriverAttributes changes in a separate PR, if you think it might help to have the fix merged faster...

Issue #453, opened 2017-01-16
keyboard


VasilijSviridov:
After running test for about 2 hours, noticed that sending keys to inputs doesn't properly works, as example at 85min 45 sec, "text" string with POST command "element/value" (also tried /keys) was posted, on UI only "t" was visible.

Any ideas why ?

VasilijSviridov:
How often do I have to restart WDA ?

marekcirkos:
@VasilijSviridov This is know XCTest issue, sometimes it fails to type strings regardless of how long it is running. Good news is that WDA can detect that and returns an error in that case, so that you can try to handle it on client side.

mykola-mokhnach:
People also say, that sometimes disabling the "Connect Hardware Keyboard" option helps to workaround this problem in Simulator.

VasilijSviridov:
Thank you, will try to play with error from WDA if that happens again.

Cheers

VasilijSviridov:
yes I am getting error you mentioned: 
Error Domain=XCTestManagerErrorDomain Code=14 \"Timed out waiting for key event to complete\" UserInfo={NSLocalizedDescription=Timed out waiting for key event to complete

VasilijSviridov:
It seems that best idea is to restart WDA after a test.

Cheers

marekcirkos:
Also if WDA is inactive for a while after test it might a good idea to keep poking it via /wda/healthcheck endpoint.

codeskyblue:
for a while means how long?

marekcirkos:
It might be a good idea to poke WDA every 20s-60s, whatever works best for you.

codeskyblue:
@marekcirkos  After call `GET /wda/healthcheck` It went to Home Screen.

I'm using this version

```
commit 2eefc210ab3a775cfd1eff1b47e1a81484b64b50
Date:   Fri Jan 13 08:51:00 2017 -0800
```

marekcirkos:
@codeskyblue This is expected. WDA tries to run some commands on XCTest to make sure connections are still there. You suppose to use this method between tests, so that it will not influence test themselves

codeskyblue:
thanks a lot

ptsiakos77:
@marekcirkos 
I run the tests in real devices there fore I cannot apply the fix with the keyboard

The thinks is that when the test hung on setValue it hangs for about 4 minutes
Is there way to reduce this timeout? (via a desired capability maybe?)

[...]

PR #655, opened 2017-07-18
Use native calls for iOS 11, since the bug with coordinates in landscape mode has been addressed there

mykola-mokhnach:
It looks like Apple finally did the job and taps in landscape mode are working properly now. However changing the device orientation still has no effect on Simulator window and it stays in portrait mode, but nobody cares.

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBTap.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBTap.m
@@ -19,6 +19,7 @@
 #import "XCElementSnapshot+FBHitPoint.h"
 #import "XCPointerEventPath.h"
 #import "XCTRunnerDaemonSession.h"
+#import "XCUICoordinate.h"
 
 const CGFloat FBTapDuration = 0.01f;
 
@@ -26,6 +27,10 @@ const CGFloat FBTapDuration = 0.01f;
 
 - (BOOL)fb_tapWithError:(NSError **)error
 {
+  if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(@"11.0")) {
+    [self tap];
+    return YES;
+  }
   XCElementSnapshot *snapshot = self.fb_lastSnapshot;
   CGPoint hitpoint = snapshot.fb_hitPoint;
   if (CGPointEqualToPoint(hitpoint, CGPointMake(-1, -1))) {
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
I've also noticed, that calling of the current tap implementation in iOS11 creates a huge delay (about 5 seconds for each tap), since XCTest has to re-read many attributes from the tree.

marekcirkos:

mykola-mokhnach:
The PR is unrelated to these fixes. There was a known bug in XCTest since 10.0 about coordinates in landscape mode were inverted. Now in iOS 11 this looks good and native _tap_ call does its job in landscape mode.

WebDriverAgentLib/Categories/XCUIElement+FBTap.m
```diff
@@ -36,6 +41,11 @@ - (BOOL)fb_tapWithError:(NSError **)error
 
 - (BOOL)fb_tapCoordinate:(CGPoint)relativeCoordinate error:(NSError **)error
 {
+  if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(@"11.0")) {
+    XCUICoordinate *tapCoordinate = [[XCUICoordinate alloc] initWithElement:self normalizedOffset:CGVectorMake(relativeCoordinate.x, relativeCoordinate.y)];
+    [tapCoordinate tap];
```
marekcirkos: Can we try not inverting coordinates and using same code path instead?

WebDriverAgentLib/Categories/XCUIElement+FBTyping.m
```diff
@@ -21,6 +22,10 @@ - (BOOL)fb_typeText:(NSString *)text error:(NSError **)error
   if (!self.hasKeyboardFocus && ![self fb_tapWithError:error]) {
     return NO;
   }
+  if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(@"11.0")) {
```
marekcirkos: Why this is needed?

mykola-mokhnach: I hope this will solve unexpected freezes described in https://github.com/facebook/WebDriverAgent/issues/481 and https://github.com/facebook/WebDriverAgent/issues/453 
The general idea is that the public _typeText_ method is fixed in XCTest after Xcode9 and there is no need to use undocumented features implemented by FBKeyboard anymore.

WebDriverAgentLib/Categories/XCUIElement+FBTap.m
mykola-mokhnach: Ok, I'll change that. My point is that usage of public API is preferred over the private one (and events synthesiser used here is the private API) and we will be happy to get rid of the workarounds as soon as we stop supporting iOS9/10 at some point of time.

mykola-mokhnach: It looks like coordinates inversion is still mandatory in iOS 11 if we use the current approach to perform tap action. Which means this PR has no point anymore. Shall I close it then?

PR #658, opened 2017-07-18
Read XCUIElement attributes directly from the instance since iOS 11

mykola-mokhnach:
The PR solves iOS 11 issues, which happen on iOS11 because of limited snapshots functionality. XCUIElement snapshot does not contain valid values even after **resolve** selector is invoked until the corresponding attribute has been explicitly read by calling _XCUIElementAttributes_ protocol property on XCUIElement instance. This makes snapshots pretty useless in iOS11. 
Also, there is one more thing to mention - each invocation of _XCUIElementAttributes_ protocol property creates a new call to the UI and slows down testing performance if the amount of such calls is  high. That is why I try to cache received values inside method calls where it is possible.

The PR is the basis for fixing xpath and predicate lookup in iOS 11.


marekcirkos:

mykola-mokhnach:
hacks,hacks,hacks... Yes, this is not needed now after snapshots logic is put back, but at some point of time it will be mandatory to apply these changes.

PR #660, opened 2017-07-20
Use dedicated screenshoting method for iOS 11

mykola-mokhnach:
The screenshot implementation based on XCAXClient_iOS is kind'a hacky and sometimes unstable (so we need to retry several times on Appium side or fall back to simctl in case of Simulator). And since iOS11 Apple finally added native support for making screenshots then why not to use it?

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIDevice+FBHelpers.m
+++ b/WebDriverAgentLib/Categories/XCUIDevice+FBHelpers.m
@@ -16,6 +16,11 @@
 #import "FBSpringboardApplication.h"
 
 #import "XCAXClient_iOS.h"
+#if defined(__has_include)
+  #if __has_include("XCTest/XCUIScreen.h")
+    #import "XCTest/XCUIScreen.h"
+  #endif
+#endif
 
 static const NSTimeInterval FBHomeButtonCoolOffTime = 1.;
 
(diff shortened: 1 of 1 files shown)


WebDriverAgentLib/Categories/XCUIDevice+FBHelpers.m
```diff
@@ -38,6 +44,17 @@ - (BOOL)fb_goToHomescreenWithError:(NSError **)error
 
 - (NSData *)fb_screenshot
 {
+  NSData *result = nil;
+  #if defined(__has_include)
+    #if __has_include("XCTest/XCUIScreen.h")
+  if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(@"11.0")) {
+    result = [[[XCUIScreen mainScreen] screenshot] PNGRepresentation];
```
marekcirkos: Can you simply check for `screenshot` selector here?

mykola-mokhnach: do you have an example? I'm not an expert in llvm preprocessing, but this was the single approach, that worked for both Xcode8 and Xcode9. The main problem there is that the whole _XCTest/XCUIScreen.h_ header file was not present prior to Xcode9. Also, if I try to call this method on iOS version < 11 with Xcode9 then it returns a valid PNG image, but the content of it is just an empty screen.

marekcirkos:
Thank you for the fix, I landing slightly simplified version.

mykola-mokhnach:
This might not work as expected in iOS 9/10 running with Xcode9. All the necessary headers are present there, but the actual method returns an empty image instead of the actual screenshot for me. That is why my PR also had system version verification. Did it work for you @marekcirkos ?

marekcirkos:
Yeah this actually wasn't great approach so I made a patch that checks for this at run time https://github.com/facebook/WebDriverAgent/commit/3979b4e178fa33e36e4e5b91a086bdfc8843dc45

PR #663, opened 2017-07-21
Fix testInvisibleDescendantWithXPathQuery

marekcirkos:


mykola-mokhnach:
It would be also useful to cache results of multiple queries to XCUIElement attributes where possible to avoid unnecessary snapshot resolutions and the related performance loss.

PR #666, opened 2017-07-24
Use backspace-delete instead of just backspace for clear

imurchie:
We have had issues where the cursor begins in the middle of the text, so just entering a backspace never finishes.

Here instead we enter a backspace (`\u0008`, moving to the left of the cursor) and a delete (`\u007F`, moving to the right of the cursor) together, so the string is deleted in both directions.

The weird machinations is to combat the error "universal character name refers to a control character" which came from trying to enter the unicode sequence directly, take [from here](https://stackoverflow.com/a/17155938).

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBTyping.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBTyping.m
@@ -30,11 +30,13 @@
 - (BOOL)fb_clearTextWithError:(NSError **)error
 {
   NSUInteger preClearTextLength = 0;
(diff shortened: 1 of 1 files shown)


marekcirkos:
Haha, cool idea. I wonder if it would be possible to send "select all" sequence and than trigger delete.
Would be way faster. XCTest for OSX allows key modifiers, it would be cool we we could trick it to work on iOS.

marekcirkos:
Tho `testTextClearing` seems to be broken

imurchie:
The test seems to work locally.

PR #687, opened 2017-08-14
Add a possibility to run multiple applications in scope of a single session

mykola-mokhnach:
It is sometimes necessary to interact with some helper apps while main application is running, for example Settings or Calendar. Xcode9 SDK introduces a possibility to handle multiple apps in scope of a single test session, activate, terminate and switch between them. 
Also, I've added a possibility to stop the application under test and then start it when needed, which might be useful for functional testing.


mykola-mokhnach:
I'm splitting this PR into two: the first one - for 'activate' endpoint and the second one - for multiple apps support.

PR #695, opened 2017-08-25
Update deprecated openurl method

mykola-mokhnach:
`- (BOOL)openURL:(NSURL*)url` is marked as deprecated since iOS 10. So we use the updated one whenever possible.


mykola-mokhnach:
Method call does not work properly together with XCTest. Not sure why, but no application is opened.

PR #702, opened 2017-09-05
Improve xpath lookup performance by analyzing the query string

mykola-mokhnach:
Reading element attributes is very expensive operation from performance point of view. The current algorithm requires to read all attributes for each accessibility element, which is going to be included into the xml document, where xpath lookup is going to be performed. This patch adds a bit more selective logic to this process, so we only read attributes, which are actually present in the client query.

I tried the patch with my local framework and performance improvement is sometimes more than 3x depending on the actual elements count in the tree (especially when we don't need to explicitly use **visible** attribute, which is the most expensive one to calculate, in the query).


SmithaVemuri:
waiting for this to be merged.

marekcirkos:
xpath should Die! ;)

Issue #705, opened 2017-09-06
Incorrect work with coordinates in landscape mode while executing press + moveTo commands


tmytlovych:
## The problem

While executing the following command in landscape, appium treats 'x' as 'y' and vise versa.

## **Environment**

Appium version (or git revision) that exhibits the issue: Appium v1.7.0-beta, v1.6.6-beta.4
Last Appium version that did not exhibit the issue (if applicable): Appium v1.6.3
Desktop OS/version used to run Appium: Sierra Version 10.12.5
Mobile platform/version under test: iPad Air 2, iOS 11.0
Real device or emulator/simulator: simulator

## **Details**

Also by executing the commands added, element is moved from the bottom of the screen by the diagonal (not just by one of the axis)

## **Appium logs**

Found matching gesture: drag

[debug] [JSONWP Proxy] Proxying [POST /wda/dragfromtoforduration] to [POST http://localhost:8100/session/1EB033B5-BDAD-4132-950A-D1B91D8C0437/wda/dragfromtoforduration] with body: {"fromX":716,"fromY":209,"toX":1543,"toY":418,"duration":1}

## **Code To Reproduce Issue**

new TouchAction(appiumDriver).press(716, 209).waitAction(1000).moveTo(827, 209).release().perform();

marekcirkos:
@mykola-mokhnach Is this Appium issue or your coordinate swapping?

mykola-mokhnach:
No, it started to happen after Xcode update to version 9. I've added a [PR](https://github.com/facebook/WebDriverAgent/pull/714) a while ago to fix that and this PR has been even already merged into our local forked WDA.

marekcirkos:
Merged. @tmytlovych Can you confirm it works for you?

quannd307:
Hi @mykola-mokhnach , about this issue [https://github.com/appium/appium/issues/9329](url), I've checked out the latest release of WDA (v2.58.1) but this issue still happens. Can you confirm that issue is solved?
Thank in advance!

marekcirkos:
@mykola-mokhnach Is it https://github.com/facebook/WebDriverAgent/issues/856 related?

mykola-mokhnach:
All of them are related if we are talking about non-default orientation.

PR #714, opened 2017-09-13
Handle coordinates calculation for non-default screen orienation, which has been fixed since iphoneos SDK 11.0

mykola-mokhnach:
XCTest has a bug when tap coordinates are inverted if device orientation differs from portrait, which has been fixed since SDK version 11.0 (Xcode9+). This PR checks the actual SDK version used to compile WDA to determine whether to apply coordinates calculation workaround.

### DIFF
--- a/WebDriverAgentTests/UnitTests/FBSDKVersionTests.m
+++ b/WebDriverAgentTests/UnitTests/FBSDKVersionTests.m
@@ -1,0 +1,67 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree. An additional grant
+ * of patent rights can be found in the PATENTS file in the same directory.
+ */
+
+#import <XCTest/XCTest.h>
+
+#import "FBRuntimeUtils.h"
+
+@interface FBSDKVersionTests : XCTestCase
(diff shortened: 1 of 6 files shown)


mykola-mokhnach:
Linking https://github.com/facebook/WebDriverAgent/issues/705

cosminstirbu:
Hello,

We'd really appreciate if this can get reviewed and merged as it is causing a high number of our tests to fail.

Thank you,
Cosmin

mykola-mokhnach:
I checked the general functionality with integration tests we already have for _fb_tap_ method. Would you like to cover other methods with unit tests?

WebDriverAgentTests/UnitTests/FBSDKVersionTests.m
```diff
@@ -0,0 +1,67 @@
...
+@property (nonatomic, readonly) NSString *higherSDKVersion;
+@end
+
+@implementation FBSDKVersionTests
+
+- (void)setUp
+{
+  [super setUp];
+  NSDictionary *bundleDict = [[NSBundle mainBundle] infoDictionary];
+  [bundleDict setValue:@"11.0" forKey:@"DTSDKName"];
```
marekcirkos: Hacky like hell :).
Easier approach might be having two layer API, one where you pass current version explicitly (that one you test) and second convenience one that passes current SDK version.

Tho we can merge this one and fix it if there is a problem.

PR #726, opened 2017-09-20
Make sure source type is set

mykola-mokhnach:
It might be that resulting error message contains "(null)" string if _format_ request argument is not set in the current implementation. This PR fixes this problem by assuring that _sourceType_ is always set to a valid string.


mykola-mokhnach:
Closed in favour of https://github.com/facebook/WebDriverAgent/pull/731

Issue #463, opened 2017-01-25
Action endpoints


marekcirkos:
Instead of implementing per gesture endpoints. Let's implement WD action endpoints that allow custom gestures. https://w3c.github.io/webdriver/webdriver-spec.html#actions

qmfrederik:
@marekcirkos I may give implementing part of the actions API a try in the coming days, in particular for touch gestures.

Before I do, I just wanted to make sure no other work is being done in this area. Also, did you have any specific design in mind for the actions API?

marekcirkos:
@qmfrederik We are not actively working on this, so feel free to give it try. We will be happy to review it.

PR #740, opened 2017-10-05
Implement support for gesture chains

mykola-mokhnach:
Implements touch actions support for Appium's and [W3C](https://github.com/jlipps/simple-wd-spec#perform-actions) protocols. The Appium's protocol might be considered as [temporary](https://github.com/appium/appium/issues/9579), since all the current versions of Appium client libs support it, which means we can start to use it immediately having minimal changes in the server code.

### DIFF
--- a/WebDriverAgent.xcodeproj/project.pbxproj
+++ b/WebDriverAgent.xcodeproj/project.pbxproj
@@ -10,6 +10,11 @@
 		711084441DA3AA7500F913D6 /* FBXPath.h in Headers */ = {isa = PBXBuildFile; fileRef = 711084421DA3AA7500F913D6 /* FBXPath.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		711084451DA3AA7500F913D6 /* FBXPath.m in Sources */ = {isa = PBXBuildFile; fileRef = 711084431DA3AA7500F913D6 /* FBXPath.m */; };
 		7119E1EC1E891F8600D0B125 /* FBPickerWheelSelectTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 7119E1EB1E891F8600D0B125 /* FBPickerWheelSelectTests.m */; };
+		71241D781FAE31F100B9559F /* FBAppiumActionsSynthesizer.m in Sources */ = {isa = PBXBuildFile; fileRef = 71241D771FAE31F100B9559F /* FBAppiumActionsSynthesizer.m */; };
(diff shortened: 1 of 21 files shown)


WebDriverAgentTests/IntegrationTests/FBTouchActionIntegrationTests.m
```diff
@@ -0,0 +1,388 @@
...
+            }
+        },
+      @{
+        @"action": @"cancel"
+        },
+      ],
+    
+    // Chain with unknown action
+    @[@{
+        @"action": @"tapP",
```
KazuCocoa: Does this mean `tap`?

mykola-mokhnach: this is intended to fail - the test verifies expected errors in case of invalid chains

marekcirkos:
I haven't seen the code in detail yet, however we could push this to kill all action endpoints and have proper WD action endpoint implementation right? https://github.com/facebook/WebDriverAgent/issues/463

mykola-mokhnach:
There are, actually, several important concerns here:
- The current implementation is compatible to Appium's TouchAction protocol, which is currently used by the clients. This a bit different from what W3C [currently defines](https://github.com/jlipps/simple-wd-spec#perform-actions), but will allow us to incorporate it with minimal effort. That is why I put the calls to new endpoints under `/wda`methods subset.
- There are too many possible chain variants. I tried to cover at least the most common ones, but there still might be issues, especially if we are talking about multiple different iOS versions/SDK support, landscape/portrait mode, etc. That is why we had an idea to merge this PR after it is reviewed by an expert to Appium's beta fork and let testers to check it first/gather the feedback.

mykola-mokhnach:
@marekcirkos I did some improvements there, so now there are endpoints for both implementations: W3C and Appium. Please take a look when you have time, because the change is quite big %)

mykola-mokhnach:
The PR is going to be merged to Appium fork of WDA

jmei08:
@mykola-mokhnach when will you merge this PR to wda please?Our team use wda only without appium.

mykola-mokhnach:
Sorry, but it's too much work for me to maintain both forks. Perhaps, we'll test the implementation with Appium first to make sure there are no major issues and then I'll create a new PR (not sure how long it would take to review it though)

Kitty2014:
@mykola-mokhnach did this release or beta？ I can help you for the Testing in our Team,thanks

PR #745, opened 2017-10-07
Use the native 'activate' method for application activation

mykola-mokhnach:
Since Xcode9 SDK Apple has added support for multiple applications in scope of a single session. This PR adds _activate_ endpoint to XCUIApplication instances and simplifies/makes more stable the implementation of _deactivateApp_ feature. This is going to be used as the base for the further PR, which adds multiple applications support to WDA.

### DIFF
--- a/WebDriverAgentLib/FBSpringboardApplication.h
+++ b/WebDriverAgentLib/FBSpringboardApplication.h
@@ -11,7 +11,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-static NSString *const SPRINGBOARD_BUNDLE_ID = @"com.apple.springboard";
+/*! Bundle identifier of Springboard app */
+extern NSString *const SPRINGBOARD_BUNDLE_ID;
 
 @interface FBSpringboardApplication : FBApplication
 
--- a/WebDriverAgentLib/FBSpringboardApplication.h
+++ b/WebDriverAgentLib/FBSpringboardApplication.h
@@ -9,9 +9,11 @@
 
 #import <WebDriverAgentLib/FBApplication.h>
 
-@interface FBSpringboardApplication : FBApplication
-
 NS_ASSUME_NONNULL_BEGIN
+
+static NSString *const SPRINGBOARD_BUNDLE_ID = @"com.apple.springboard";
+
+@interface FBSpringboardApplication : FBApplication
 
 /**
  @return FBApplication that is attached to SpringBoard
--- a/WebDriverAgentLib/Utilities/FBXCodeCompatibility.m
+++ b/WebDriverAgentLib/Utilities/FBXCodeCompatibility.m
@@ -26,8 +26,13 @@ static dispatch_once_t onceRootElementToken;
 
 @end
 
+
+NSString *const FBApplicationMethodNotSupportedException = @"FBApplicationMethodNotSupportedException";
+
 static BOOL FBShouldUseOldAppWithPIDSelector = NO;
 static dispatch_once_t onceAppWithPIDToken;
+static BOOL FBCanUseActivate = NO;
+static dispatch_once_t onceActivate;
 @implementation XCUIApplication (FBCompatibility)
 
 + (instancetype)fb_applicationWithPID:(pid_t)processID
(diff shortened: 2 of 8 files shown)


WebDriverAgentLib/FBSpringboardApplication.h
```diff
@@ -9,10 +9,12 @@
 
 #import <WebDriverAgentLib/FBApplication.h>
 
-@interface FBSpringboardApplication : FBApplication
-
 NS_ASSUME_NONNULL_BEGIN
 
+static NSString *const SPRINGBOARD_BUNDLE_ID = @"com.apple.springboard";
```
marekcirkos: Nice! However can you move it to '.m' file? This will complain about redefinition if you include this header twice.
If you need it outside, probably you don't, you can always add in header this
```
extern static NSString *const SPRINGBOARD_BUNDLE_ID;
```

mykola-mokhnach: piece of cake :)

WebDriverAgentLib/Utilities/FBXCodeCompatibility.m
```diff
@@ -41,5 +46,29 @@ + (instancetype)fb_applicationWithPID:(pid_t)processID
...
+  });
+  if (!FBCanUseActivate) {
+    [[NSException exceptionWithName:FBApplicationMethodNotSupportedException reason:@"'activate' method is not supported by the current iOS SDK" userInfo:@{}] raise];
+  }
+  [self activate];
+}
+
+- (NSUInteger)fb_state
+{
+  return [[self valueForKey:@"state"] intValue];
```
marekcirkos: What was the point of removing state from XCUIApplicaiton?

WebDriverAgentTests/UnitTests/Doubles/FBApplicationDouble.h
```diff
@@ -11,4 +11,5 @@
 
 @interface FBApplicationDouble : NSObject
 @property (nonatomic, assign, readonly) BOOL didTerminate;
+@property (nonatomic) NSString* bundleID;
```
marekcirkos: Please add explicit `strong`

WebDriverAgentTests/UnitTests/Doubles/FBApplicationDouble.m
```diff
@@ -15,6 +15,15 @@ @interface FBApplicationDouble ()
 
 @implementation FBApplicationDouble
 
+- (instancetype)init
+{
+  self = [super init];
+  if (self) {
+    self.bundleID = @"some.bundle.identifier";
```
marekcirkos:
```
_bundleID = @"some.bundle.identifier";
```

WebDriverAgentLib/Utilities/FBXCodeCompatibility.m
mykola-mokhnach: the signature of that method has changed since Xcode9, so the compiler was complaining about such inconsistence.

WebDriverAgentLib/Utilities/FBXCodeCompatibility.m
```diff
@@ -41,5 +46,29 @@ + (instancetype)fb_applicationWithPID:(pid_t)processID
   return [self applicationWithPID:processID];
 }
 
+- (void)fb_activate
+{
+  dispatch_once(&onceActivate, ^{
```
marekcirkos: You have implemented same thing twice. You probably want to use `fb_isActivateSupported` here

WebDriverAgentTests/UnitTests/Doubles/FBApplicationDouble.h
mykola-mokhnach: added

WebDriverAgentTests/UnitTests/Doubles/FBApplicationDouble.m
mykola-mokhnach: changed

WebDriverAgentLib/Utilities/FBXCodeCompatibility.m
mykola-mokhnach: fixed, thanks

WebDriverAgentLib/Utilities/FBXCodeCompatibility.m
marekcirkos: How different? They changed `NSUInteger` to `NSInteger` ?

PR #749, opened 2017-10-10
add remote screen and control command 

butterflyingdog:
as someone use webdriveragent to get captured screen of real device and control the device from web page, but the processing of default tap and drag commands is slow slow( about 3-5 seconds), I use XCEventGenerator to speed  the processing.

### DIFF
--- a/WebDriverAgent.xcodeproj/project.pbxproj
+++ b/WebDriverAgent.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
I would start with code polishing in order to make it comply with general style guideline (remove comments, unnecessary settings, format code blocks properly).
The second step would be to add unit/integration tests.

qmfrederik:
@butterflyingdog If you want to perform arbitrary mouse actions using the WebDriver protocol, an option could be to implement the [actions routes](https://www.w3.org/TR/webdriver/#actions) instead of adding a custom one.

I had a stab once at implementing the validation logic of the actions route - see https://github.com/qmfrederik/WebDriverAgent/commit/053ae1648c2d18f097bba22ca194a200193695fb, but never got the time to finish it.

PR #758, opened 2017-10-24
add remote access and control command

butterflyingdog:
As someone use WebDriverAgent through web browser to get captured screen of real device and
control the device from web page, but the processing of default tap and
drag commands is slow slow( about 3-5 seconds), I use XCEventGenerator
to speed the processing.


butterflyingdog:
I found that the test cases causing Travis CI build failed  isn't my added code, why?

PR #764, opened 2017-10-26
Exclude time-intensive checks from visibility verification

mykola-mokhnach:
_hitPoint_ calculation is super-time intensive and it's even more time-intensive if we do it for all the descendant elements, which can greatly slow down source generation, because it's repeated for each node. My pure assumption is that

```objc
CGPoint midPoint = [self.suggestedHitpoints.lastObject CGPointValue];
   XCElementSnapshot *hitElement = [self hitTest:midPoint];
   if (self == hitElement || [self._allDescendants.copy containsObject:hitElement]) {
     return YES;
   }
```

verification should also cover the two conditions, which were removed in this PR.

I'd like to ask other people to test this change.

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
@@ -17,7 +17,6 @@
 #import "XCUIElement+FBUtilities.h"
 #import "XCTestPrivateSymbols.h"
 #import <XCTest/XCUIDevice.h>
-#import "XCElementSnapshot+FBHitPoint.h"
 
 @implementation XCUIElement (FBIsVisible)
 
@@ -49,14 +48,6 @@
   XCElementSnapshot *hitElement = [self hitTest:midPoint];
   if (self == hitElement || [self._allDescendants.copy containsObject:hitElement]) {
     return YES;
-  }
-  if (CGRectContainsPoint(appFrame, self.fb_hitPoint)) {
-    return YES;
-  }
-  for (XCElementSnapshot *elementSnapshot in self._allDescendants.copy) {
-    if (CGRectContainsPoint(appFrame, elementSnapshot.fb_hitPoint)) {
-      return YES;
-    }
   }
   return NO;
 }
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
@@ -17,6 +17,7 @@
 #import "XCUIElement+FBUtilities.h"
 #import "XCTestPrivateSymbols.h"
 #import <XCTest/XCUIDevice.h>
+#import "XCElementSnapshot+FBHitPoint.h"
 
 @implementation XCUIElement (FBIsVisible)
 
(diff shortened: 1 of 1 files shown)


antiarchit:
`hitPoint` is not always the mid point. Also, if we compute `hitPoint`, we don't need to do `hitTest` I believe. I'm looking into redoing some of the visibility stuff since it seems to be the rate limiting step. If you have any ideas, do let me know.

mykola-mokhnach:
For me it looks like _hitPoint_ verification is only needed for Xcode SDK below 9 and on the most recent SDK _hitTest_ with _suggestedHitpoints_ works as expected. But I need https://github.com/facebook/WebDriverAgent/pull/714 to be merged first to be able to distinguish the current SDK version.

antiarchit:
I'm trying to address this on my end as well. It seems like just calculating `hitPoint` does it for us now? 

```diff
@@ -45,20 +45,7 @@
   if (!CGRectIntersectsRect(frame, screenFrame)) {
     return NO;
   }
-  CGPoint midPoint = [self.suggestedHitpoints.lastObject CGPointValue];
-  XCElementSnapshot *hitElement = [self hitTest:midPoint];
-  if (self == hitElement || [self._allDescendants.copy containsObject:hitElement]) {
-    return YES;
-  }
-  if (CGRectContainsPoint(appFrame, self.fb_hitPoint)) {
-    return YES;
-  }
-  for (XCElementSnapshot *elementSnapshot in self._allDescendants.copy) {
-    if (CGRectContainsPoint(appFrame, elementSnapshot.fb_hitPoint)) {
-      return YES;
-    }
-  }
-  return NO;
+  return CGRectContainsPoint(appFrame, self.fb_hitPoint);
 }
```

What do you think?

mykola-mokhnach:
I assume it was like this before you did the commit 16253f7. 

I still like the idea of splitting the implementation for Xcode 9 and higher and Xcode8.
Also, it would be interesting to know which call is less expensive - hitPoint or hitTest?

antiarchit:
> I assume it was like this before you did the commit 16253f7.

Yes, that is correct. I think just the hit point check should be fine. 

> I still like the idea of splitting the implementation for Xcode 9 and higher and Xcode8.

I think we're at full Xcode 9 support now so it's fine if stuff doesn't work with Xcode 8. However, we should take care if stuff breaks on iOS 10 or lower. 

> Also, it would be interesting to know which call is less expensive - hitPoint or hitTest?

I'd have to experiment on multiple runs. The time I added the check on descendants it did improve the visibility checks to be more accurate. But maybe that's not the case since Xcode 9/iOS 11 versus Xcode 8/iOS 11.

mykola-mokhnach:
> I think we're at full Xcode 9 support now so it's fine if stuff doesn't work with Xcode 8. However, we should take care if stuff breaks on iOS 10 or lower.

According to my previous experience iOS version does not play big role there, but rather SDK version. At least on Simulator. Means _hitTest_ works on iOS10 and 11 with Xcode9 SDK, but does not work properly on iOS10 with Xcode8 SDK (and Travis run, actually, confirms that: https://travis-ci.org/facebook/WebDriverAgent/jobs/293139167).

mykola-mokhnach:
BTW, Xcode8 does not support iOS11 ;)

antiarchit:
WDA built against Xcode 8 works on iOS 11 afaik. 

Also, I  checked and realised we need both for now. There are cases when `hitPoint` is wrong.

mykola-mokhnach:
Closed in favour of https://github.com/facebook/WebDriverAgent/pull/794

PR #776, opened 2017-11-02
Add support for multiple applications with Xcode9 SDK

mykola-mokhnach:
Xcode9 introduced support for multiple applications running in scope of a single XCTest session. This PR adds several new endpoints in order to make this functionality available from WDA


marekcirkos:
What is the use case here? You will not be able to handle e.g. `findElement` queries cross different application right?

mykola-mokhnach:
for example you have an application, which has a separate entry in system preferences. This allows to open preferences, change the stuff there and then return back to the app. 
Also, it makes possible to implement test cases, where the application under test needs to be restarted

shynkevich-alex:
@marekcirkos Could you merge it?

mykola-mokhnach:
The PR has been merged to Appium form of WDA

LaforgeN:
Hi @mykola-mokhnach , we are using Appium with Python. Was this implemented as well? If so, wou;d it be possible to share how to switch between two apps using Python client.

PR #783, opened 2017-11-07
Use correct property to detect interface orientation

mykola-mokhnach:
Experiments approve that `[XCUIDevice sharedDevice].orientation` always returns portrait orientation even if the actual device orientation has been changed to landscape, however `FBApplication.fb_activeApplication.interfaceOrientation` works as expected.


mykola-mokhnach:
Will reopen the PR from the other branch

Issue #746, opened 2017-10-09
iOS 11 displayed value returns false incorrectly for XCUIElementTypeImage


rajfidel:
## The problem

We have an iOS test application.
When installed on a device with iOS 10.x, the displayed value for XCUIElementTypeImage elements is "true" (which is correct)
When installed on a device with iOS 11.0, the displayed value for XCUIElementTypeImage elements is "false" (incorrect)
I can see that the image is visible on the GUI on checking manually in both cases.
This issue is only present on **XCUIElementTypeImage** elements. (Easily reproducible by checking the image displayed property value on iOS Settings application com.apple.Preferences)

Both devices are iPhone 6s (same form factor),
Appium 1.7.0 + Xcode 9 is used in both cases above.
WebdriverAgent was removed and freshly installed on the iPhone in both cases.
There are no overlapping GUI components that overlap with the XCUIElementTypeImage element as per the page source (Not sure if it matters, but checked anyways, just in case)

On debugging WebdriverAgent:
On iOS 11.0 - hitPoint is -1.000000, -1.000000 (incorrect)
On iOS 10.x - hitPoint is 47.500000, 291.500000 (correct) which causes the below code snippet in XCUIElement+FBIsVisible.m to execute (in iOS 10.x) or not execute (in iOS 11.0)
```
  if (CGRectContainsPoint(appFrame, self.fb_hitPoint)) {
    return YES;
  }
```

## Environment

* Appium version (or git revision) that exhibits the issue: 1.7.0
* Last Appium version that did not exhibit the issue (if applicable): n/a
* Desktop OS/version used to run Appium: mac OS 10.12.6
* Node.js version (unless using Appium.app|exe): 8.5.0
* Mobile platform/version under test: iOS 11.0
* Real device or emulator/simulator: real device
* Appium CLI or Appium.app|exe: Appium CLI

## Details

Got the displayed property and the Page source for the following configurations:

Appium 1.7.0 + iOS 10 + Xcode 9 --> Page source [here](https://gist.github.com/rajfidel/e633c1f7e0801c4dacb4585bcd9e05e7), Appium logs [here](https://gist.github.com/rajfidel/c35aae43d5359859340c10e1cb1e0233)
[...]

PR #794, opened 2017-11-16
A new word in visibility detection

mykola-mokhnach:
This PR speeds up visibility detection in general, fixes it for landscape mode and also fixes scrolling to element for Xcode9+
Previously visibility detection was based on _hittable_ property value, but this property does not work for static elements, like an image in a cell, since hittablity for such elements is defined by gesture recogniser. Now we use another approach: we detect the visibility of the particular element based on the intersection between its own frame and all ancestor frames. This gives better results, since even if the element is not hittable the viability value is calculated properly for it:
before
now

As a result, the issues like https://github.com/facebook/WebDriverAgent/issues/746 should be addresses after the PR is merged and the overall visibility detection speed should be better.

### DIFF
--- a/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
+++ b/WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
@@ -27,19 +27,19 @@
 
 @implementation XCElementSnapshot (FBIsVisible)
 
-+ (BOOL)fb_isVisibleInContainerWindow:(XCElementSnapshot *)window selfElement:(XCElementSnapshot *)element hierarchyIntersection:(nullable NSValue *)intersectionRectange appFrame:(CGRect)appFrame
+- (BOOL)fb_isVisibleInContainerWindow:(XCElementSnapshot *)window hierarchyIntersection:(nullable NSValue *)intersectionRectange appFrame:(CGRect)appFrame
 {
-  CGRect currentRectangle = nil == intersectionRectange ? element.frame : [intersectionRectange CGRectValue];
-  XCElementSnapshot *parent = element.parent;
+  CGRect currentRectangle = nil == intersectionRectange ? self.frame : [intersectionRectange CGRectValue];
+  XCElementSnapshot *parent = self.parent;
   CGRect intersectionWithParent = CGRectIntersection(currentRectangle, parent.frame);
   if (CGRectIsEmpty(intersectionWithParent)) {
     return NO;
   }
   if (parent == window) {
     // Assume the element is visible if its root container is hittable and the frame is onscreen
-    return CGRectContainsPoint(appFrame, element.fb_hitPoint);
+    return CGRectContainsPoint(appFrame, self.fb_hitPoint);
   }
-  return [self.class fb_isVisibleInContainerWindow:window selfElement:parent hierarchyIntersection:[NSValue valueWithCGRect:intersectionWithParent] appFrame:appFrame];
+  return [parent fb_isVisibleInContainerWindow:window hierarchyIntersection:[NSValue valueWithCGRect:intersectionWithParent] appFrame:appFrame];
 }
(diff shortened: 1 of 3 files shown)


mykola-mokhnach:
@antiarchit Scrolling has been fixes, because I've actually noticed this bug while running visibility integration tests on my machine with Xcode9:
> And regarding the scrolling issue: Xcode9 always puts nil into self.application property of all XCElementSnapshot instances. That is why we need to explicitly pass the current application instance from the corresponding XCUIElement instance, where it is defined.

But yep, it would be better to extract this fix into a separate PR

WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
```diff
@@ -30,35 +27,36 @@ - (BOOL)fb_isVisible
 
 @implementation XCElementSnapshot (FBIsVisible)
 
++ (BOOL)fb_isVisibleInContainerWindow:(XCElementSnapshot *)window selfElement:(XCElementSnapshot *)element hierarchyIntersection:(nullable NSValue *)irect appFrame:(CGRect)appFrame
+{
+  XCElementSnapshot *parent = element.parent;
+  CGRect intersectionRect = nil == irect ?
```
antiarchit: `irect ?` should do it?

WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
```diff
@@ -30,35 +27,36 @@ - (BOOL)fb_isVisible
 
 @implementation XCElementSnapshot (FBIsVisible)
 
++ (BOOL)fb_isVisibleInContainerWindow:(XCElementSnapshot *)window selfElement:(XCElementSnapshot *)element hierarchyIntersection:(nullable NSValue *)irect appFrame:(CGRect)appFrame
```
antiarchit: `intersectionRectangle` instead?

mykola-mokhnach:
@antiarchit nope, scrolling changes were moved to https://github.com/facebook/WebDriverAgent/pull/806

WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
mykola-mokhnach: renamed

WebDriverAgentLib/Categories/XCUIElement+FBIsVisible.m
mykola-mokhnach: renamed

antiarchit:
Still showing up in the [commit](https://github.com/facebook/WebDriverAgent/pull/794/commits/dfe0bb60ab5376546f8ba0910de4fb25f58f1ee3).

Also, I wonder if `fb_isVisibleInContainerWindow:selfElement:hierarchyIntersection:appFrame:` would be better as an instance method on `XCElementSnapshot`. The parameter `selfElement` seems extraneous.

mykola-mokhnach:
No problems, I'll squash it. Regarding the instance method - it is not the case here, as the `element` value is different on different recursion level and contains ancestor elements of the original one

antiarchit:
> Regarding the instance method - it is not the case here, as the element value is different on different recursion level and contains ancestor elements of the original one

Sure, but you can call it on the parent after the parent is computed?

mykola-mokhnach:
> Sure, but you can call it on the parent after the parent is computed?

Perhaps then I'll have to replace recursion with a loop. Would you prefer this approach?

antiarchit:
> Perhaps then I'll have to replace recursion with a loop. Would you prefer such approach?

The same recursion would work. You'll be calling the method on `parent` instead of `self.class`, that's all.

mykola-mokhnach:
Updated. Thanks for the hint

PR #804, opened 2017-11-29
Improve XML performance

chendo:
Rather than instantiating an object for every attribute written, this PR turns the `FBElementAttribute` `recordWithWriter` calls into class methods instead.

On the main screen of our application, dumping the XML tree went from 3.5s~ to 2.4s~, for a performance increase of 30%~.

### DIFF
--- a/WebDriverAgentLib/Utilities/FBXPath.m
+++ b/WebDriverAgentLib/Utilities/FBXPath.m
@@ -23,10 +23,9 @@
 @property (nonatomic, readonly) id<FBElement> element;
 
 + (nonnull NSString *)name;
-- (nullable NSString *)value;
++ (nullable NSString *)valueForElement:(id<FBElement>)element;
 
-- (instancetype)initWithElement:(id<FBElement>)element;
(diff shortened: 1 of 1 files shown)


WebDriverAgentLib/Utilities/FBXPath.m
```diff
@@ -531,25 +531,23 @@ + (NSString *)name
...
 + (NSString *)name
 {
   return kXMLIndexPathKey;
 }
 
-- (NSString *)value
++ (int)recordWithWriter:(xmlTextWriterPtr)writer forValue:(NSString *)value
 {
-  return self.indexValue;
+  if (nil == value) {
```
mykola-mokhnach: this value can never be equal to null. Also, the returned value here is always expected to be unique for each node

lavi-xu:
I just tied this patch on 10.3.3 with xcode 8.3.3. 

WOW. This works very perfect.

Thank you, @chendo 

BRs.

pluie-cloud:
@chendo @facebook-github-bot  hi I find XCTestWD use AEXML to process source,it's very fast,almost 1s ,Does OC has this frame?

PR #806, opened 2017-11-30
Fix scrolling under Xcode9

mykola-mokhnach:
Xcode9 always puts `nil` into self.application property of all XCElementSnapshot instances. That is why we need to explicitly pass the current application instance from the corresponding XCUIElement instance, where it is defined.


marekcirkos:
Omg....

PR #867, opened 2018-02-15
Fix remote bridged views detection in xpath source

mykola-mokhnach:
This is the backport of fixes we did for Appium in order to properly display RemoteBridgeView  hierarchies in page source and be able to find descendant element there using xpath queries in Xcode 9+. 

The original problem source somehow lies in XCUIApplication snapshots. If the snapshot is taken from any other parent element (we use direct children of the application element) then all its descendant remote views are properly visible in the hierarchy.

The PR also applies some refactoring on page source displaying logic in general.

### DIFF
--- a/WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
+++ b/WebDriverAgentLib/Categories/XCElementSnapshot+FBHelpers.m
@@ -33,7 +33,7 @@ inline static BOOL isSnapshotTypeAmongstGivenTypes(XCElementSnapshot* snapshot, 
 
 - (NSArray<XCElementSnapshot *> *)fb_descendantsMatchingXPathQuery:(NSString *)xpathQuery
 {
-  return (NSArray<XCElementSnapshot *> *)[FBXPath findMatchesIn:self xpathQuery:xpathQuery];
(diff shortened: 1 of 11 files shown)


marekcirkos:
I think I solved XCUApplication problem with https://github.com/facebook/WebDriverAgent/commit/183f679ac0908c266ceb8286c86186eb8d2be995. Will check that diff tomorrow. Can you just remove white space changes?

sridharreddy123:
Eagerly waiting for the fix to merge!!

zastress:
Any chance that this will be merged soon?

zastress:
I am not able to upload files from iCloud until this is merged... I am guessing there are others with the same problem. Merging this would be a major improvement :).

marekcirkos:
@mykola-mokhnach Thank you for your work and sorry for late response on this. However you touch so much stuff that I do not feel comfortable merging it eg.
- fb_lastSnapshot
- source tree repsonse
- xpath implementation

Can you split this diff and explain a bit more?

mykola-mokhnach:
Yes, there are too many changes here and it would be even more complicated to backport this change from Appium's branch, because it's too different now. Perhaps I'll find some time for it in the future

PR #870, opened 2018-02-22
Fix for exception when serializing element dictionaries with INFINITE…

JonGabilondoAngulo:
… values. Targets dependencies on RoutingHTTPSever.framework fixed to get WDARunner signature correctly.

Found at least a case of elements with INFINITE values in x,y. In Springboard widgets screen, at the bottom there are some "Link" elements with INFINITE x,y. This breaks the serialisation of the elements dictionary.

On the other hand, a more general problem is that RoutingHTTPServer.framework is not code signed because it is embedded into the WebRiverAgentLib framework. This is a documented problem and the solution is to make the Runner know that the framework exists.


WebDriverAgentLib/Categories/XCUIElement+FBWebDriverAttributes.m
```diff
@@ -161,4 +161,15 @@ - (NSDictionary *)wdRect
   };
 }
 
+- (NSDictionary *)wdRectNoInf
+{
+  CGRect frame = self.wdFrame;
+  return @{
+           @"x": @((frame.origin.x != INFINITY) ?frame.origin.x :0),
```
mykola-mokhnach: such values are returned by XCTest and I don't think it makes a lot of sense to build such workarounds over them. I'd rather try to find the reason of why it happens. Perhaps, it is some wrong way of reading element attributes in WDA source or a bug on the Apple side?

JonGabilondoAngulo: My debugging says it's an Apple bug. When the Links are not visible, at least the ones at the bottom of the Springboard, they show the INFINITE values. I need the springboard tree, so this patch helps.

mykola-mokhnach: why not to use `isinf`?

WebDriverAgent.xcodeproj/project.pbxproj
```diff
@@ -2079,10 +2092,12 @@
 			baseConfigurationReference = EEE5CABF1C80361500CBBDD9 /* ProjectSettings.xcconfig */;
 			buildSettings = {
 				CLANG_STATIC_ANALYZER_MODE = deep;
+				CODE_SIGN_IDENTITY = "iPhone Developer";
```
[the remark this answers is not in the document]
JonGabilondoAngulo: It shouldn't sorry.

WebDriverAgentLib/Categories/XCUIElement+FBWebDriverAttributes.m
JonGabilondoAngulo: Nicer with isinf, thanks.

PR #871, opened 2018-02-22
INFINITY values avoided in element dictionary. Runner frameworks code…

JonGabilondoAngulo:
And Runner frameworks codesign fixed.

Some elements may have INFINITY values in their frame rect, causing an exception in the serialisation of the element dictionary. This has been found in Link elements o the Springboard (Widgets page, at the bottom, links like Weather or Stocks copyright info).

The runner can't  be Run on devices directly from Xcode due to RoutingHTTPServer framework not being code signed. This pull request has a fix in the project build steps.


antiarchit:
Can you please separate out the changes in two different pull requests since you're dealing with two different issues?

PR #873, opened 2018-02-23
Codesigning RoutingHTTPServer.framework when building Runner.

JonGabilondoAngulo:
Running the Runner from Xcode fails because RoutingHTTPRunner is not code signed. Being embedded into WebDriverAgentLib, Xcode doesn't know to sign it.
This PR adds a build step insert a signed RoutingHTTPServer.framework into WebDriverAgentLib.

### DIFF
--- a/WebDriverAgent.xcodeproj/project.pbxproj
+++ b/WebDriverAgent.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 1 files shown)


marekcirkos:
Why this is needed? We already have "Code Sign on Copy" for this. 
If you want to run WDA on device you just need to specify project team for provisioning profiles.

JonGabilondoAngulo:
From my Xcode 9.3, when running in test mode to install and run the wda runner, it complains that the embedded framework is not code signed. I think sign on copy does not sign embedded ones.

marekcirkos:
Already mearged

Issue #856, opened 2018-01-22
A Short Explanation: Why there are problems with landscape orientation


mykola-mokhnach:
Facebook decided to close Wiki from editing by third-parties, so I just leave it here.

## Bug Description

There are many complains about interaction issues in non-default screen orientation mode. All these problems are coming from a single source. XCTest has a known bug when it fails to transform element coordinates properly in non-portrait orientation. It means that if https://github.com/facebook/WebDriverAgent/blob/52512a7fb39a697fc727d372c8765a43c74f6d37/WebDriverAgentLib/Commands/FBOrientationCommands.m#L101 was called with orientation value different from `UIDeviceOrientationPortrait` then interface elements, which belong to `some` windows from the current hierarchy, will change their frames, while others won't. 

## The Problem

Usually, it is the case, that if the particular window belongs to a system service then its descendants will have "correct" frame coordinates and if it a custom application then the4 coordinates will be "corrupted". Under "corrupted" I assume that `click` or other actions will be executed not in the place, where we expect it to be executed (or simply offscreen) if sent to such element.  This also includes the case when coordinates are calculated  based on the global screen size, since the "real" screen dimensions constructed by XCTest are not the same as returned to you.

The described problem also affects, for example, screenshots. The are incorrectly rotated in non-portrait mode (the bottom side of the picture is not the bottom of the screen).

Also, it is known, that the problem shows itself differently in iOS9 and iOS10+.

## The Solution
The WDA implementation from Facebook tries to workaround the problem by adjusting coordinates based on the current interface orientation value. It assumes the coordinates should be always fixed if the orientation is different from the default one. See 

https://github.com/facebook/WebDriverAgent/blob/52512a7fb39a697fc727d372c8765a43c74f6d37/WebDriverAgentLib/Commands/FBElementCommands.m#L457

and 

https://github.com/facebook/WebDriverAgent/blob/52512a7fb39a697fc727d372c8765a43c74f6d37/WebDriverAgentLib/Categories/XCUIElement%2BFBTap.m#L37

Here you have the methods, which "fix" the coordinates: 

https://github.com/facebook/WebDriverAgent/blob/52512a7fb39a697fc727d372c8765a43c74f6d37/WebDriverAgentLib/Utilities/FBMathUtils.m#L41
https://github.com/facebook/WebDriverAgent/blob/52512a7fb39a697fc727d372c8765a43c74f6d37/WebDriverAgentLib/Utilities/FBMathUtils.m#L56

[...]

Issue #798, opened 2017-11-20
The /wda/tap endpoint not working correctly in landscape mode


rpes:
My env: 
  iPhone 6s with iOS 11.1.2, Xcode 9.1 (9B55)
  WebDriverAgent master code from Nov 20, 2017

Symptoms:

The /wda/tap end point is not performing correctly when the app is in the landscape mode. 
Taps are being applied only to a certain limit of the X-coordinate. It feels like the coordinates 
are internally filtered by the bounding box of the portrait application (375x667).

The issue is not new to the latest iOS 11.1.2. I had the same problem about 2 months ago with an older WDA version and the initial iOS 11 release.

Steps to reproduce:
1. Start the iBooks application
2. Rotate the device to the landscape left mode
3. Select the "My Books" screen (see the screen shot)
4. Tap "Most Recent" at [180,50] -> the tap performs correctly
5. Tap "Authors" - the tap gets applied only if x<375
6. No taps with x>=375 are possible - no way to tap for example "Categories" or "Select"

![ibooks](https://user-images.githubusercontent.com/5822023/33031064-6d5cf2f0-ce1d-11e7-9efd-b684f6684845.png)

The log:

$ export JSON_HEADER='-H "Content-Type: application/json"'

$ curl -X POST $JSONHEADER -d "{\"desiredCapabilities\":{\"bundleId\":\"com.apple.iBooks\"}}" http://192.168.1.8:8100/session
{
  "value" : {
    "sessionId" : "1D07CE20-34DA-4F86-9B66-FFC96739C9AD",
    "capabilities" : {
      "device" : "iphone",
      "browserName" : "iBooks",
      "sdkVersion" : "11.1.2",
      "CFBundleIdentifier" : "com.apple.iBooks"
    }
  },
  "sessionId" : "1D07CE20-34DA-4F86-9B66-FFC96739C9AD",
  "status" : 0
}

Tap “Most Recent” - SUCCESSFUL

$ curl -X POST $JSONHEADER -d "{\"x\":180,\"y\":50}" http://192.168.1.8:8100/session/1D07CE20-34DA-4F86-9B66-FFC96739C9AD/wda/tap/0
{
  "value" : {

  },
  "sessionId" : "1D07CE20-34DA-4F86-9B66-FFC96739C9AD",
  "status" : 0
}

Tap “Categories” - UNSUCCESSFUL, TAP NOT APPLIED

[...]

Issue #732, opened 2017-09-26
Appium 1.7.0 + iOS10.x not finding element on new app if manually navigated to new app


rajfidel:
## The problem

This below mentioned issue is only present in iOS 10.x + Appium 1.7.0:
1. If appium session was started for a bundleID (in this case com.fidlee.ICMHelper), and another app is manually opened (in this case com.apple.Preferences), findElement does not find any element present on com.apple.Preferences.
PageSource shows all the elements correctly though.
2. FindElement only works if the appium session was started for the bundleID on which the element is present. So if appium session was started for bundleID com.apple.Preferences, it would be able to find elements present in com.apple.Preferences application.

This issue was not present in Appium 1.6.5 on iOS 10.x
On Appium 1.7.0 this issue is only present on iOS 10.x. 
**Appium 1.7.0 with iOS 11.0 works fine**.

Appium Logs are available below:
1. On iOS 10 + Appium 1.7.0, create session for com.fidlee.ICMHelper, manually launch com.apple.Preferences and try to find an element. Logs available [here](https://gist.github.com/rajfidel/a64124d5399d26167d5db79782133dc7) (**Is unable to find element** All other scenarios described below work fine)
2. On iOS 10 + Appium 1.7.0, create session for com.apple.Preferences and try to find an element. Logs available [here](https://gist.github.com/rajfidel/145120d2c8e7c31c872ab6632723b510) (Is able to find element)
3. On iOS 11 + Appium 1.7.0, create session for com.fidlee.ICMHelper, manually launch com.apple.Preferences and try to find an elements. Logs available [here](https://gist.github.com/rajfidel/9b9c624a553fe20b32c9d6a5a4ade74c) (Is able to find element)
4. On iOS 11 + Appium 1.7.0, create session for com.apple.Preferences and try to find an element. Logs available [here](https://gist.github.com/rajfidel/ac7c2320f39bd91722f6b0970118e188) (Is able to find element)

## Environment

* Appium version (or git revision) that exhibits the issue: 1.7.0
* Last Appium version that did not exhibit the issue (if applicable): 1.6.5
* Desktop OS/version used to run Appium: mac OS 10.12.6
* Node.js version (unless using Appium.app|exe): 8.5.0
* Mobile platform/version under test: 10.3.1
* Real device or emulator/simulator: Real device
* Appium CLI or Appium.app|exe: Appium CLI

marekcirkos:
[...]

PR #878, opened 2018-03-01
Fix the incorrect tap coordinate in landscape mode.

DeviLeo:
Summary:
  If SDK >= 11, the tap coordinate based on application is not correct when
  the application orientation is landscape and
  tapX > application portrait width or tapY > application portrait height.
  Pass the window element to the method [FBElementCommands gestureCoordinateWithCoordinate:element:] will resolve the problem.
  More details about the bug, please see the following issues: #705, #798, #856.
  Notice: On iOS 10, if the application is not launched by wda, no elements will be found. See issue #732.

### DIFF
--- a/WebDriverAgentLib/Commands/FBElementCommands.m
+++ b/WebDriverAgentLib/Commands/FBElementCommands.m
@@ -460,8 +460,55 @@ static const CGFloat DEFAULT_OFFSET = (CGFloat)0.2;
   if (shouldApplyOrientationWorkaround) {
     point = FBInvertPointForApplication(coordinate, application.frame.size, application.interfaceOrientation);
   }
-  XCUICoordinate *appCoordinate = [[XCUICoordinate alloc] initWithElement:application normalizedOffset:CGVectorMake(0, 0)];
-  return [[XCUICoordinate alloc] initWithCoordinate:appCoordinate pointsOffset:CGVectorMake(point.x, point.y)];
+  XCUIElement *element = application;
+  if (isSDKVersionGreaterThanOrEqualTo(@"11.0")) {
+    XCUIElement *window = [self findWindowInApplication:application];
+    if (window) element = window;
+  }
+  return [self gestureCoordinateWithCoordinate:point element:element];
+}
+
+/**
+ Returns gesture coordinate based on the specified element.
+ 
+ @param coordinate absolute coordinates based on the element
+ @param element the element in the current application under test
+ @return translated gesture coordinates ready to be passed to XCUICoordinate methods
+ */
++ (XCUICoordinate *)gestureCoordinateWithCoordinate:(CGPoint)coordinate element:(XCUIElement *)element
+{
+  XCUICoordinate *appCoordinate = [[XCUICoordinate alloc] initWithElement:element normalizedOffset:CGVectorMake(0, 0)];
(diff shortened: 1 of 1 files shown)


WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -460,8 +460,55 @@ + (XCUICoordinate *)gestureCoordinateWithCoordinate:(CGPoint)coordinate applicat
...
+ will resolve the problem.
+ More details about the bug, please see the following issues:
+ #705: https://github.com/facebook/WebDriverAgent/issues/705
+ #798: https://github.com/facebook/WebDriverAgent/issues/798
+ #856: https://github.com/facebook/WebDriverAgent/issues/856
+ Notice: On iOS 10, if the application is not launched by wda, no elements will be found.
+ See issue #732: https://github.com/facebook/WebDriverAgent/issues/732
+ */
++ (XCUIElement *)findWindowInApplication:(XCUIApplication *)application {
+  NSArray<XCUIElement *> *allElements = application.windows.allElementsBoundByIndex;
```
mykola-mokhnach: this code is redundant, since only the first window is needed. You may use `fb_firstMatch` call

mykola-mokhnach: `
application.windows.fb_firstMatch
`

WebDriverAgentLib/Commands/FBElementCommands.m
```diff
@@ -460,8 +460,55 @@ + (XCUICoordinate *)gestureCoordinateWithCoordinate:(CGPoint)coordinate applicat
   if (shouldApplyOrientationWorkaround) {
     point = FBInvertPointForApplication(coordinate, application.frame.size, application.interfaceOrientation);
   }
-  XCUICoordinate *appCoordinate = [[XCUICoordinate alloc] initWithElement:application normalizedOffset:CGVectorMake(0, 0)];
-  return [[XCUICoordinate alloc] initWithCoordinate:appCoordinate pointsOffset:CGVectorMake(point.x, point.y)];
+  XCUIElement *element = application;
+  if (isSDKVersionGreaterThanOrEqualTo(@"11.0")) {
```
mykola-mokhnach: method signature should be changed if `shouldApplyOrientationWorkaround ` parameter is not used anymore

DeviLeo: For backwards compatibility reason, I think ```shouldApplyOrientationWorkaround``` is useful if SDK < 11.

JonGabilondoAngulo:
Your are right on this one but there is one important thing that cause bad calculations. Using "firstMatch" window can get a window that is bigger(different) than the device screen, or the window where the tap must be done. 
I had to search in all windows and get the window with the same size as the App size.
The fix I'm using successfully is based on making sure the window is has the correct size.
If you could check cases where this happens .. I can't tell u right now where a found such a case. I could in some days from now. Let me know.

mykola-mokhnach:
@JonGabilondoAngulo That is why I always advice people to avoid clicking by absolute coordinates. It is much more reliable to click on an element, since you may calculate whether coordinates for it are inverted or not and this is not possible if one provides absolute screen coordinates.

However the current solution might be good as a poor-man's one. I'd also add `interfaceOrientation` verification, so we only apply it if it is not the portrait one.

akhand-pratap2807:
Still I am getting this issue. Not able to tap on correct coordinate.

PR #882, opened 2018-03-06
Return correct responses if xpath or class chain locator is invalid

mykola-mokhnach:
This should improve elements lookup flow if done by native clients by returning correct webdriver response codes for incorrect locator expressions.


marekcirkos:
@mykola-mokhnach Can you email me? I lost your email :)

mykola-mokhnach:
Yep, thanks for the suggestion

PR #897, opened 2018-03-27
Use a natural UUID for elements in the element cache

qmfrederik:
Every time an element is added to the cache, a new UUID is generated for that element.

If the same element is resolved multiple times (though the same or different queries), multiple copies of that element persist in the cache.

The "multiple copies" problem exists because the WebDriverAgent doesn't directly access the UI elements (which live in another process) but uses proxies/remoting/... to invoke those elements, and you end up with multiples proxies for the same element.

It appears, however, it is still possible to determine whether two `XCUIElement` objects point to the same element by looking at element -> snapshot -> accessibilityElement -> elementId + processIdentifier.

You can then avoid duplicates in the element cache by generating a UUID based on this information. That's what this PR does.

A simple test which queries the same element 1000 times now causes the WebDriverAgent to consume 23MB on the device instead of 126MB, with memory usage fairly stable during the test.

Might not be worth the effort if we can implement a LRU-like cache but at least wanted to get feedback on the general idea.

### DIFF
--- a/PrivateHeaders/XCTest/XCAccessibilityElement.h
+++ b/PrivateHeaders/XCTest/XCAccessibilityElement.h
@@ -13,11 +13,15 @@
     int _processIdentifier;
     struct __AXUIElement *_axElement;
     unsigned long long _elementType;
+    unsigned long long _elementOrHash;
+    unsigned long long _elementID;
 }
 @property(readonly) id payload; // @synthesize payload=_payload;
 @property(readonly) int processIdentifier; // @synthesize processIdentifier=_processIdentifier;
 @property(readonly) const struct __AXUIElement *AXUIElement; // @synthesize AXUIElement=_axElement;
 @property(readonly, getter=isNative) BOOL native;
+@property(readonly) unsigned long long elementID; // @synthesize elementID=_elementID;
+@property(readonly) unsigned long long elementOrHash; // @synthesize elementOrHash=_elementOrHash;
 
 + (id)elementWithAXUIElement:(struct __AXUIElement *)arg1;
 + (id)elementWithProcessIdentifier:(int)arg1;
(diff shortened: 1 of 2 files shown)


qmfrederik:
@mykola-mokhnach What do you think? Worth the effort or too risky?

mykola-mokhnach:
Would wdUID property be ok for that purpose?

qmfrederik:
It seems very similar 😄.

I was just wondering - do you know whether `_elementId` is unique in the context of a device, or just an individual process?

I used both the process ID and the element ID to conflicts when the same element ID is assigned for an element in two different processes, but not sure whether this is needed.

mykola-mokhnach:
yes these ids are unique in scope of a single process

mykola-mokhnach:
however this algorithm might be more expensive from performance perspective, since you need to resolve element snapshot each time you save it into the ache

qmfrederik:
In that case, I would use both the process ID + wdUID for calculating the UUID.
Regarding cost of the algorithm, I haven't observed a noticeable delay in my tests (but let me run some numbers tomorrow); plus you might argue in most cases the user will interact with the element after resolving it, so the cost will be incurred anyway.

But I agree there are trade-offs.

mykola-mokhnach:
process id can be taken from element's `application` property

mykola-mokhnach:
@qmfrederik What do you think about https://github.com/appium/WebDriverAgent/pull/65?

qmfrederik:
Looks good to me, I'll comment there.

marekcirkos:
I understand that this PR is dropped?

qmfrederik:
@marekcirkos Well, are you interested in porting the work done in appium/WebDriverAgent#65 over to the Facebook repository? It uses "natural" identifiers for elements, and greatly reduces the size of the element cache if users repeatedly query for the same element.

I can create a PR for that.

Same goes for appium/WebDriverAgent#66 which implements a LRU cache. I'd be happy to port those changes to the Facebook repository if there's interest on your side to land them.

marekcirkos:
@qmfrederik since there is new Xcode work coming. I am pretty sure that part of our work will be on improving WDA's performance, so we might end up with something similar, but slightly different.

In general I plan to limit number of magic stuff we are doing as XCTest framework is getting better and better.

PR #903, opened 2018-04-04
Test fail 

pluie-cloud:
os 10.13.4 
IOS IPAD 11.3
when i test webdriveragent-runner  test fail


marekcirkos:
This is like project sync diff?

Issue #902, opened 2018-04-04
Xcode 9.3 Unable to launch WebDriverAgent because of xcodebuild failure: xcodebuild failed with code 65


MalcolmCNZ:
Hi There

Webdriver agent was working fine without any Issues, But when i updated to xcode 9.3, i get the"xcode failure" error message, Just wondering if i am the only one who is getting this message after updating to xcode 9.3... All the provisioning certificates are signed.. Please advise.

Thanks
Malcolm

AnuKadloor:
Hi @MalcolmCNZ ,
Even I'm facing this issue. I'm using iOS 11.3, xcode 9.3, appium 1.7
The tests were working fine in xcode 9.2
@imurchie Can you help?

Below is the error:-
Error Domain=NSCocoaErrorDomain Code=260 "The file “WebDriverAgentRunner-Runner.app” couldn’t be opened because there is no such file." UserInfo={NSFilePath=/Users/admin/Library/Developer/Xcode/DerivedData/WebDriverAgent-brdadhpuduowllgivnnvuygpwhzy/Build/Products/Debug-iphoneos/WebDriverAgentRunner-Runner.app, NSUnderlyingError=0x7f8ed32950a0 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}
[Xcode] 
[Xcode] 2018-04-04 16:30:24.875 xcodebuild[5369:174509] Error Domain=IDETestOperationsObserverErrorDomain Code=6 "Early unexpected exit, operation never finished bootstrapping - no restart will be attempted" UserInfo={NSLocalizedDescription=Early unexpected exit, operation never finished bootstrapping - no restart will be attempted, NSUnderlyingError=0x7f8ed32a7a00 {Error Domain=NSCocoaErrorDomain Code=260 "The file “WebDriverAgentRunner-Runner.app” couldn’t be opened because there is no such file." UserInfo={NSFilePath=/Users/admin/Library/Developer/Xcode/DerivedData/WebDriverAgent-brdadhpuduowllgivnnvuygpwhzy/Build/Products/Debug-iphoneos/WebDriverAgentRunner-Runner.app, NSUnderlyingError=0x7f8ed32950a0 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}}}}
[Xcode] 
[debug] [JSONWP Proxy] Proxying [GET /status] to [GET http://localhost:8100/status] with no body
[Xcode] 
[...]

PR #915, opened 2018-04-18
Include RoutingHTTPServer in copy to fix ISSUE #902

antonkushpil:
Include RoutingHTTPServer in copy to fix ISSUE #902

### DIFF
--- a/WebDriverAgent.xcodeproj/project.pbxproj
+++ b/WebDriverAgent.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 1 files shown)


NickAb:
Once build I just removed the copy of RoutingHTTPServer framework form `WebDriverAgentRunner-Runner.app/PlugIns/WebDriverAgentRunner.xctest/Frameworks/RoutingHTTPServer.framework` and was still able to run tests using WebDriverAgent successful.

Removing RoutingHTTPServer.framework from `WebDriverAgentRunner.xctest/Frameworks` and keeping it only in `Frameworks/WebDriverAgentLib.framework/Frameworks` saves 4Mb.

@antonkushpil could you explain why is it needed and if it is needed during build phase only?

PR #917, opened 2018-04-25
force touch support

alexanderbalaban:
A couple of commands which make possible to force touch given element or point by coordinates. In fact – copy of the tap extension. Baked with integration tests.

Updated with @mykola-mokhnach improvements in  appium#79


marekcirkos:
This is awesome! Thank you for this

PR #964, opened 2018-07-18
Added support to get source for WKWebView

sumanc:
Hi,

When I use the following command;
curl -X GET $JSON_HEADER $DEVICE_UL/session/7D046797-5D22-4E16-8264-754C5A1C4796/source

and if the app has a WKWebView, I am not getting any elements from the WKWebView.

I couldn't find any place else other than this pull request. With this change, I am able to get the complete source of the app including the WKWebView.

Please let me know if this is the right thing todo or if there is a better way.

--Suman


sumanc:
I an try. I just looked at the project's structure. I need to add a WKWebview based view controller to the integration and add a test case in the integration tests?

Please send me if there are any guidelines to add a test case.

zastress:
I had the same issue with Webviews and this commit fixed it. Thanks @sumanc .

PR #976, opened 2018-08-11
Fix port value selection

mykola-mokhnach:
It turns out, that `NSProcessInfo.processInfo.environment[@"USE_PORT"]` returns an empty string instead of nil if `USE_PORT` is not present in the environment which causes zero value to be assigned by default.


marekcirkos:
Facepalm. Thank you for working on this! :)

marekcirkos:
@mykola-mokhnach Could you add test case to `FBConfigurationTests`?

Issue #975, opened 2018-08-10
Xcode Beta 5 iOS 12 Unable to create WebDriverAgent session with any application


fr0l:
Dear FB team and project maintainers,

When trying to create a new WebDriver session for MobileSafari (as well as any other app) then WebDriverAgent responds with the following error (see full stacktrace below) :

```
<FBApplicationProcessProxy 0x6000016809e0>
    addObserver:<_XCKVOExpectationImplementation 0x6000035a74d0>
    forKeyPath:@\"isProcessIDValid\"
    options:7
    context:0x6000035a74d0]

was sent to an object that is not KVC-compliant for the \"isProcessIDValid\" property.
```
 
Command to start the WebDriver session:
```
curl -X POST '-H "Content-Type: application/json"' -d '{"desiredCapabilities":{"bundleId":"com.apple.mobilesafari"}}' http://localhost:42799/session
```

Please could you help to troubleshoot / fix the issue? (trying to run our tests on iOS 12)

Details that might help:

Environment:
```
Xcode: Version 10.0 beta 5 (10L221o)
MacOS X: High Sierra 10.13.6

WebDriverAgent: c4da863b9c8d6d076ba74bf1ff58b47ef56dd2fa
    Build locally with Xcode 10.0 beta 5 (10L221o)

Fbsimctl: f2faff8d5503c4f028928094431328935d6ea56a
    Build locally with Xcode 10.0 beta 5 (10L221o)
    
iOS Simulator: 'iPhone 8' 'iOS 12'
```

WebDriverAgent was started by following command:
```
/usr/local/bin/fbsimctl --json $SIM \
    launch_xctest $WDA/PlugIns/WebDriverAgentRunner.xctest \
    com.apple.MobileAddressBook \
    --port 42799 \
    -- listen
```

```
{
  "value" : "
  [

  <FBApplicationProcessProxy 0x6000016809e0>
      addObserver:<_XCKVOExpectationImplementation 0x6000035a74d0>
      forKeyPath:@\"isProcessIDValid\"
      options:7
      context:0x6000035a74d0]

  was sent to an object that is not KVC-compliant for the \"isProcessIDValid\" property.


(
	0   CoreFoundation                      0x00000001105ad73b __exceptionPreprocess + 331
	1   libobjc.A.dylib                     0x000000010fb4a735 objc_exception_throw + 48
	2   CoreFoundation                      0x00000001105ad595 +[NSException raise:format:] + 197
[...]

PR #977, opened 2018-08-13
Forward NSKeyValueObserverRegistration methods to proxied object

fr0l:
Add forwarding of NSKeyValueObserverRegistration methods.

Fixes issue #975 (Forward KVC-compliant methods to proxied object)

### DIFF
--- a/WebDriverAgentLib/FBApplicationProcessProxy.m
+++ b/WebDriverAgentLib/FBApplicationProcessProxy.m
@@ -25,6 +25,8 @@
   return proxy;
 }
 
+#pragma mark - Override XCUIApplicationProcess methods
+
 - (void)waitForQuiescence
 {
   if (!self.shouldWaitForQuiescence) {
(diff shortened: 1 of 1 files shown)


marekcirkos:
Thank you!

fr0l:
@marekcirkos this fix raises need to implement forwarding of methods in protocol NSKeyValueCoding
:-)

fr0l:
It looks like better approach would be to subclass from NSProxy . I'll try that in a moment

marekcirkos:
Why close?

fr0l:
This PR opens need for need to implement some methods from ```NSKeyValueCoding```  protocol (for example ```valueForUndefinedKey```). I think probably the solution not to bodge here and there would be to use ```NSProxy``` as base class for ```FBApplicationProcessProxy```

but that would require a bit more code. but it's a more generic solution.

sorry I was too quick to open this PR (facepalm)

PR #978, opened 2018-08-13
Use NSProxy as base class for FBApplicationProcessProxy 

fr0l:
Allows to forward NSKeyValueObserverRegistration methods.
Fixes issue #975


fr0l:
Hey @marekcirkos  
Please could you take a look at this PR. It's the ```NSProxy``` variant?
the better implementation than in closed PR #977 

I've tested this with iOS 12. now testing with iOS 11.4

marekcirkos:
Why close?

fr0l:
Hey @marekcirkos 
sorry. i should have just push the fix to the PR, but decided it's better close and perform a bit more careful investigation.

The PR itself is fine. but there are situations, when in a single test one could create and delete several WebDriver sessions. 
Basically in ```FBApplication``` when performing ```[applicationProcess isMemberOfClass:XCUIApplicationProcess.class]``` 
would return ```YES``` as it would be forwarded by the proxy (```NSProxy```) to the proxied object (```XCUIApplicationProcess```).

To fix — I've added ```isProxy``` in ```FBApplication``` check as well ```NSParameterAssert``` in ```FBApplicationProcessProxy```

and pushed this to a new PR #979 (this one is final) (facepalm)
:-)

PR #979, opened 2018-08-15
Use NSProxy as base class for FBApplicationProcessProxy

fr0l:
Allows to forward NSKeyValueObserverRegistration methods. 

Fixes issue #975 — Allows to work with Xcode 10 beta 5 and beta 6.

Tested for Xcode 9.4.1 as well

### DIFF
--- a/WebDriverAgentTests/UnitTests/FBApplicationProcessProxyTests.m
+++ b/WebDriverAgentTests/UnitTests/FBApplicationProcessProxyTests.m
@@ -25,7 +25,7 @@
 @end
 
 @interface FBApplicationProcessProxy (ProxiedMethod)
-- (NSInteger)proxiedMethod;
+- (int)proxiedMethod;
 @end
 
 @interface XCUIApplicationProcess (TestableMethods)


WebDriverAgentLib/FBApplication.m
```diff
@@ -122,7 +122,7 @@ - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(N
     return;
   }
   XCUIApplicationProcess *applicationProcess = change[NSKeyValueChangeNewKey];
-  if (!applicationProcess || ![applicationProcess isMemberOfClass:XCUIApplicationProcess.class]) {
+  if (!applicationProcess || [applicationProcess isProxy] || ![applicationProcess isMemberOfClass:XCUIApplicationProcess.class]) {
```
marekcirkos: I don't think that explicitly checking for proxy here is what we want, as we shouldn't be even thinking about it being a proxy. `isMemberOfClass` should be also forwarded to `XCUIApplicationProcess` and would return YES , which is what we want right? So the `if` statement would behave correctly here.

WebDriverAgentLib/FBApplicationProcessProxy.m
```diff
@@ -20,11 +20,18 @@ @implementation FBApplicationProcessProxy
...
   NSParameterAssert(applicationProcess);
-  FBApplicationProcessProxy *proxy = [self.class new];
+  NSParameterAssert([[applicationProcess class] isEqual:XCUIApplicationProcess.class]);
+  FBApplicationProcessProxy *proxy = [[self.class alloc] init];
   proxy.applicationProcess = applicationProcess;
   return proxy;
 }
 
+- (instancetype)init {
+  return self;
```
marekcirkos: What is the point of init here?

WebDriverAgentLib/FBApplicationProcessProxy.m
```diff
@@ -45,9 +52,16 @@ - (void)waitForQuiescenceIncludingAnimationsIdle:(BOOL)includeAnimations
   [self.applicationProcess waitForQuiescenceIncludingAnimationsIdle:includeAnimations];
 }
 
-- (id)forwardingTargetForSelector:(SEL)aSelector
```
marekcirkos: Why kill `forwardingTargetForSelector`? it is faster then invocation right?

WebDriverAgentTests/UnitTests/FBApplicationProcessProxyTests.m
```diff
@@ -0,0 +1,82 @@
...
+  [exception raise];
+}
+
+@end
+
+@interface FBApplicationProcessProxyTest : XCTestCase
+
+@end
+
+@implementation FBApplicationProcessProxyTest
```
marekcirkos: Nice!

WebDriverAgentLib/FBApplication.m
fr0l: If remove check for ```[applicationProcess isProxy]``` then execution of the method ```observeValueForKeyPath``` would continue, and the line of code
```
[object setValue:[FBApplicationProcessProxy proxyWithApplicationProcess:applicationProcess] forKey:keyPath];
```
gets executed. 

if the ```applicationProcess``` is a proxy object (```FBApplicationProcessProxy```), then it would instantiate a new proxy object (```FBApplicationProcessProxy```) with parameter ```applicationProcess``` (which is a proxy itself already).

This happens in rare cases , when in one test several applications are used (several WebDriver sessions) — for example: 1. Start session with Safari, 2. Stop session, 3. Start session with Preferences, 4. Stop session, 5. Start session with Safari, 2. Stop session

In this case it would start a recursive calls to proxied objects.

WebDriverAgentLib/FBApplicationProcessProxy.m
fr0l: I haven't done any benchmarking, and implemented the only two required methods. if this is faster, then let's use it.

WebDriverAgentLib/FBApplicationProcessProxy.m
fr0l: NSProxy does not implement ```init``` method.

WebDriverAgentLib/FBApplication.m
marekcirkos: You are absolutely right, I completely missed the point of that `if` ;P.

WebDriverAgentLib/FBApplicationProcessProxy.m
marekcirkos: From https://developer.apple.com/documentation/objectivec/nsobject/1418855-forwardingtargetforselector
```
This method gives an object a chance to redirect an unknown message sent to it before the much more expensive forwardInvocation: machinery takes over.
```

WebDriverAgentLib/FBApplicationProcessProxy.m
fr0l: Without implementing ```init``` method, there is an error trying to instantiate ```FBApplicationProxy``` object.

```
failed: caught "NSInvalidArgumentException", "*** +[FBApplicationProcessProxy init]: unrecognized selector sent to class
```

The same for the ```new``` method. Both methods ```init``` and ```new``` are implemented in ```NSOjbect``` and not in ```NSProxy```

fr0l:
@marekcirkos  Thanks for the hint on the ```forwardingTargetForSelector``` method.

I did benchmarking — basically a test which was calling ```XCTAssertEqual([proxy proxiedMethod], 1);``` 10 million times. 

And received surprising and impressive results.

```FBApplicationProcessProxy``` WITH implemented ```forwardingTargetForSelector``` | ```FBApplicationProcessProxy``` WITHOUT implemented ```forwardingTargetForSelector```
------------------------ | ------------------------
~ 0.3 seconds | ~ 26 seconds

WebDriverAgentLib/FBApplicationProcessProxy.m
fr0l: Thanks. Simple benchmark with 10 million invocations answers the question "How ```much more expensive```?" :-) 0.3 seconds vs 26 seconds

marekcirkos:
Nice work! Thanks for contributing!

jamesvanhorn:
Thanks for quick fix

